PR #310, opened 2014-12-18
Use the much more awesome pygit2.

### LINKED ISSUE #309: Port the git post receive hook to libgit2
We've had more than 1 bug report on the git bindings used by the current post receive hook.

We should probably port to something more maintained and modern.  @pypingou, how is that new lib you were working on?

ralphbean:
It is really much more awesome.

With the old code we would generate exceptions if:

* The user pushed a git tag.
* The user pushed a commit signed with a gpg key.

Furthermore, it was just mistakenly written and it would only publish
one fedmsg even if multiple commits were pushed (it would only broadcast
about the latest commit).  This code actually publishes messages for
each commit on each branch.

Fixes #309.

### DIFF
--- a/extras/git-hooks/post-receive
+++ b/extras/git-hooks/post-receive
@@ -1,16 +1,13 @@
 #!/usr/bin/env python
 
 import getpass
-import git
-import itertools
 import os
 import sys
 
+import pygit2
+
 import fedmsg
 import fedmsg.config
-
-# Read in all the rev information git-receive-pack hands us.
-lines = [line.split() for line in sys.stdin.readlines()]
 
 # Use $GIT_DIR to determine where this repo is.
 abspath = os.path.abspath(os.environ['GIT_DIR'])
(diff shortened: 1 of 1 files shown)


ralphbean:
Looking at this now, writing a test for this is going to involve more code than the actual hook does (I think).  We would need to create a bare repo, install the hook, clone the repo, commit, push, catch the message.  It's really tricky.  More tricky than it's worth, imo.

pypingou:
Could this help? https://github.com/fedora-infra/pygit2_utils/blob/master/tests/__init__.py#L44

ralphbean:
> Could this help?

Sure.  It's not importable though, right?  So -- copy it wholesale?

pypingou:
>  It's not importable though, right?

Well it is, if you depend on pygit2_utils :-)

> So -- copy it wholesale?

That might be the easiest, :+1: for me

ralphbean:
So, as to the test... even if I copied that block to create a bare repo, (add my git hook that I want to test), clone it, add files, commit them, and push.. In order to test I'd need to *catch the network activity of the git hook* which would involve installing or mocking out a fedmsg-relay.  It's too much for me.

When new releases of pygit2 come out, can we just test them in staging with a git push to our dist-git setup there?

If this fails in production due to a backwards-incompatible release of pygit2, the worst case is that a fedmsg message doesn't get published and the user gets a nasty traceback but the commit goes through just fine.  They'll complain and we can fix it.

Even if that happens many times, I sense that it will take only a fraction of the developer time that writing this test will.  (And repairing the test in the future will consume even more unknown cycles.)

pypingou:
Sounds good to me (especially if we use stg for testing first)

ralphbean:
Cool.  Thanks for all the pointers @pypingou :)

PR #313, opened 2015-02-04
Fix ValueError when using gpg and gpg_home is not set

mbooth101:
There is a simple type-o in the config key that prevents this from working.


ralphbean:
Nice catch!  Thanks :)

PR #316, opened 2015-02-10
Surface CRL errors.

ralphbean:
This should fix the bug reported here:
https://bugzilla.redhat.com/show_bug.cgi?id=1189082

If there's a network error and the CRL is not reachable, currently
fedmsg just quietly returns the local path where the CRL should be, even
though it hasn't been downloaded.  This should surface that exception to
the user of the library.


pypingou:
:+1: for me

PR #317, opened 2015-02-11
Make IRC lineRate configurable.

### LINKED ISSUE #299: make the irc bot lineRate configurable
In ``fedmsg/consumers/ircbot.py`` there is a ``lineRate`` value hardcoded at `0.6`.  This should be made configurable.

Today, freenode staff let us know that we sometimes go over our rate limit and some messages get dropped (freenode staff get notified when this happens, so it is annoying).

We should:

- At very least increase our hardcoded value to ``1.0``.
- Better, make it configurable via the config dict.

ralphbean:
Fixes #299.

### DIFF
--- a/fedmsg/consumers/ircbot.py
+++ b/fedmsg/consumers/ircbot.py
@@ -90,11 +90,12 @@ def make_irc_client(factory):
     class Fedmsg2IRCClient(irc.IRCClient):
         # The 0.6 seconds here is empircally guessed so we don't get dropped by
         # freenode.  FIXME - this should be pulled from the config.
-        lineRate = 1.0
+        lineRate = factory.rate
         sourceURL = "https://github.com/fedora-infra/fedmsg"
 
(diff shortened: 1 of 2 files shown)


fedmsg/consumers/ircbot.py
```diff
@@ -208,12 +210,13 @@ def __init__(self, hub):
             pretty = settings.get('make_pretty', False)
             terse = settings.get('make_terse', False)
             short = settings.get('make_short', False)
+            rate = settings.get('line_rate', 1.0)
```
pypingou: The default is different from what is in ircbot.py?

pypingou:
One question, otherwise :+1:

fedmsg/consumers/ircbot.py
ralphbean: It is.  I just wanted to prove to myself that I could make them different and it could work.  ;p

PR #318, opened 2015-02-17
Include constituent metadata in the conglomerate message.

ralphbean:
This should fix fedora-infra/fedmsg_meta_fedora_infrastructure#165.

### DIFF
--- a/fedmsg/meta/base.py
+++ b/fedmsg/meta/base.py
@@ -109,8 +109,21 @@ class BaseProcessor(object):
             'packages': ['ghc', 'nethack', ... ],
             'topics': ['org.fedoraproject.prod.git.receive'],
             'categories': ['git'],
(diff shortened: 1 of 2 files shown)


pypingou:
Code wise :+1: but that means we're including metadata within the fedmsg message?

Would there be no way to iterate through the bulk messages and run fedmsg.meta on them client side?

ralphbean:
> that means we're including metadata within the fedmsg message?

Yes and no.  These aren't the original fedmsg here, they're the "conglomerate" messages which get produced by ``fedmsg.meta``.  Each conglomerate messages represents a few original messages, so we're stuffing the metadata of those originals in the new one.

> Would there be no way to iterate through the bulk messages and run fedmsg.meta on them client side?

Not from Fedora Mobile (not written in python); the original idea is in fedora-infra/mobile#42.

PR #319, opened 2015-02-17
Handle multiple topics.

ralphbean:
Whoops!  Consumers can possibly declare their ``self.topic`` to be a list.
Here we get pages of results from datagrepper and only yield the messages
if the topic matches *any one* of the topics that we're interested in.

### DIFF
--- a/fedmsg/consumers/__init__.py
+++ b/fedmsg/consumers/__init__.py
@@ -187,14 +187,20 @@ class FedmsgConsumer(moksha.hub.api.consumer.Consumer):
         data = _make_query()
 
         # Grab and smash subsequent pages if there are any
+        interesting_topics = self.topic
+        if not isinstance(interesting_topics, list):
+            interesting_topics = [interesting_topics]
+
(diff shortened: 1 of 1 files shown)


fedmsg/consumers/__init__.py
```diff
@@ -185,17 +185,22 @@ def _make_query(page=1):
...
+
         for page in range(1, data['pages'] + 1):
             self.log.info("Retrieving datagrepper page %i of %i" % (
                 page, data['pages']))
             data = _make_query(page=page)
 
             for message in data['raw_messages']:
-                if message['topic'].startswith(self.topic[:-1]):
-                    yield message
+                for topic in interesting_topics:
```
pypingou: Do we want to check first if there are interesting_topics?

ralphbean: Ah, there has to be at least one or the consumer won't load and get this far.

pypingou:
:+1: thanks!

PR #321, opened 2015-02-19
Scrub args from keywords before initializing.

ralphbean:
This is wild.

The symptom you'd get is this:  The *first* time
``fedmsg.publish(topic=topic, msg=dict(...))`` is ever called, fedmsg
isn't initialized yet.  Those topic and msg arguments are passed through
to ``init(**kw)`` as extra keyword arguments and so "topic" and "msg"
appear in the global config dict (internally).

This never mattered in practice.  They were there, and noone cared.

I'm hitting it now because in fmn, we try to call
``fedmsg.meta.msg2link(msg=msg, **config)`` but there is a ``msg`` key
in config from the first time ``publish`` was called and so it fails
with "cannot have two keyword arguments named 'msg'".

### DIFF
--- a/fedmsg/__init__.py
+++ b/fedmsg/__init__.py
@@ -19,6 +19,7 @@
 #
 """ Fedora Messaging Client API """
 
+import inspect
 import threading
 
 import fedmsg.core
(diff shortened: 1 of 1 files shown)


fedmsg/__init__.py
```diff
@@ -62,9 +63,16 @@ def init(**kw):
...
     def api_function(func):
+        scrub = inspect.getargspec(func).args
 
         def _wrapper(*args, **kw):
             if not hasattr(__local, '__context'):
+                kw = kw.copy()
+
+                for arg in scrub:
+                    if arg in kw:
+                        del kw[arg]
```
pypingou: Would we want to log which arg are dropped?

ralphbean: Heh, I don't think logger isn't set up at this point.  It would need its config which hasn't been loaded yet.

pypingou:
After some discussions and thoughts on IRC, :+1: for me

PR #323, opened 2015-02-23
Avoid recursively walking the entire git history.

ralphbean:
This will run us into the limit for recursive calls as jwb just pointed out
in ``#fedora-admin``.

### DIFF
--- a/extras/git-hooks/post-receive
+++ b/extras/git-hooks/post-receive
@@ -41,22 +41,6 @@ def revs_between(head, ancestors):
         yield head.id
 
 
-def gather_ancestors(commit, seen=None):
-    """ Yield all ancestors commits of a given commit. """
-
-    seen = seen or []
-    idx = commit.id
-
-    if not idx in seen:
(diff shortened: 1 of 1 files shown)


extras/git-hooks/post-receive
```diff
@@ -97,7 +81,7 @@ for line in lines:
 
     try:
         base = repo.revparse_single(base)
-        ancestors = list(gather_ancestors(base))
+        ancestors = [commit.id for commit in repo.walk(base.id)]
```
pypingou: Should we pass it sorting info?

pypingou:
One question, otherwise :+1:

extras/git-hooks/post-receive
ralphbean: Yes, likely.

ralphbean: I take it back.  The list of ancestors can be in any order.  We're only checking for membership in it to figure out when to stop second part of the hook later on.

PR #324, opened 2015-03-10
Add more debug info to test suite failures.

ralphbean:
Currently, when tests fail it is kind of difficult to figure out exactly
*which* test failed.  This should make it much more easy to jump to the failed
test to inspect it.


pypingou:
:fireworks: :+1:

PR #325, opened 2015-03-10
Re-use the produce_template function for consistency.

ralphbean:
As it stood before, these two kinds of things had different templates:

- Actual conglomerations of multiple messages.
- Fake conglomerations of a single message.

This re-use of code ensures that they will actually look the same.


pypingou:
:+1: for me

PR #326, opened 2015-03-12
No more infinite loops.

ralphbean:
We hit an infinite loop in the git hook again tonight which made for a *lot* of
traffic on the bus.

This rolls back the functionality to what we used to have before the move to
gitolite3/rhel7.  It translates as "if someone pushes 5 commits, only publish
about the latest one".


pypingou:
Sad but ok :+1: 

Would we want to give a try to shelling out to git to retrieve the list of commits?

ralphbean:
> Would we want to give a try to shelling out to git to retrieve the list of commits?

Yeah.  After thinking about it over the weekend, this seems like the best next option.

PR #327, opened 2015-03-19
Some git hook improvements.

ralphbean:
Aimed at making it so we can disable the original scm git hook emails.


pypingou:
:+1: for me

PR #328, opened 2015-03-29
Publish git commits in chronological order.

ralphbean:
We currently publish them newest to oldest which can be really confusing if you
receive them over email.  @remicollet pointed this out in ``#fedora-devel``
last week.


ralphbean:
Note, we need to apply this to our ansible repo if/when it gets merged.

pypingou:
:+1: for me

PR #331, opened 2015-04-02
Add a tool to scan fedmsg endpoints.

ralphbean:
This will be useful after freeze for figuring out which of our endpoints need
to be cleaned out (for instance, we still have some defined for app01,app02,
etc.. which no longer exist).

With the collectd mode, we can produce a graph that we want to be at "100%"
most of the time.. but we can watch and see when it dips as things restart and
see how long it takes for things to get back to full activation.


scripts/fedmsg-map.py
```diff
@@ -0,0 +1,125 @@
...
+            connection = None
+            try:
+                connection = socket.create_connection(endpoint, timeout)
+                actual = base64.b64encode(connection.recv(10))
+                if actual != expected:
+                    inactive[name].append((
+                        endpoint, "%r is not %r" % (actual, expected)))
+                    info("F")
+                else:
+                    active[name].append((endpoint, "all active"))
```
pypingou: I understand the ``active`` but not why it says ``*all* active``, does it loop through the port or so?

pypingou:
One question, otherwise :+1: for me

scripts/fedmsg-map.py
ralphbean: Oh, yeah.  the 'all' if a typo.  It does loop through the ports, but is says "all" active for each one.  It should just say "active".

PR #332, opened 2015-04-13
Initial shot at Python 3 support

bkabrda:
This PR contains initial changes to make fedmsg support Python 3. Some notes:
- All tests are green, but I think that doesn't really mean anything :)
- There are still some `isinstance(..., basestring)` calls and since I don't have enough knowledge about fedmsg, I'm not sure what to do with these. They seem to be used as input for `json.loads`, which only takes PY3's `str` type.
- The real problem, I think, will be M2crypto, since that doesn't support Python 3 at all for now. Maybe it would make sense to move to cryptography [1] in long term, since that seems to be *the* Python crypto library for following years.

[1] https://pypi.python.org/pypi/cryptography/

### DIFF
--- a/fedmsg/commands/trigger.py
+++ b/fedmsg/commands/trigger.py
@@ -21,7 +21,7 @@ import re
 import subprocess
 import sys
 import threading
-import Queue
+from six.moves import queue
 
 import fedmsg
 import fedmsg.encoding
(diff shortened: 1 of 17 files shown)


ralphbean:
:+1:!  Thanks :)

fedmsg/commands/trigger.py
```diff
@@ -110,11 +110,11 @@ def run(self):
         max_queue_size = int(self.config['max_queue_size'])
 
         timer = None
-        queue = Queue.Queue()
+        que = queue.Queue()
```
glensc: `que` is not english word, why rename the import (comparing to original code)?
it's ugly if your code reads like you can't spell (do common spellinge errors).

`from six.moves import queue as Queue` afaik works fine

bkabrda: I decided to do this because I consider renaming imports misleading. When we finally fully switch to Python 3, there will be no `Queue` import, just `queue`.
I find it usual that people use variables named `klass` or so, so that's why I chose `que`, which sounds pretty much like `queue` when I read it aloud.

PR #333, opened 2015-04-13
Documentation Fix and Flake-8 compliance

Nolski:
I updated the docs so they list the correct directories for where it checks the directories. The collections module wasn't being used in config.py either so I removed that


ralphbean:
:+1: thanks!

PR #334, opened 2015-04-14
Make list_to_series return uniqified items in predictable order.

bkabrda:
This helps with testing dependent packages (e.g. fedmsg-meta-fedora-infrastructure) on Python 3, which has randomization of sets and maps turned on.


pypingou:
:+1: for me

PR #336, opened 2015-05-07
Make order of configuration files predictable

msimacek:
The order in which configuration files are loaded matters, but fedmsg.config
uses listdir, which returns files in order depending on the filesystem. Sorting
them by name should make things more predictable.


ralphbean:
:+1:  Thanks @msimacek!

PR #337, opened 2015-05-09
Use a threadpool to speed this up.

ralphbean:
... so we can increase the resolution of the collectd graph.


pypingou:
cool :+1:

PR #338, opened 2015-05-09
Flatten the threadpool workload.

ralphbean:
... so that there is less work per item and more threads have a greater effect.  This gets the runtime in our infra down to about 2s (from 5s).

This is a followup to #337.


lmacken:
:+1:

PR #343, opened 2015-06-24
Get a release ready with python3 support.

ralphbean:
This includes the work done by @bkabrda to port the crypto module to python3.
We found that:

- M2Crypto won't be ported to python3, so we need something else.
- python-cryptography looks nice.
- It can only do the message-signing part of what we need, not message validation.

So this pull request introduces an (ugly) split.  If we're on py3, we use an
x509_ng module that uses cryptography for signing .. and will fail to validate
anything with an ``ImportError`` on M2Crypto.  If we're on py2, then we use the
old standard x509 module.

I'd like to release this so we can try it with mailman3 which is python3 only.
We have [a plugin for it](https://github.com/fedora-infra/mailman3-fedmsg-plugin)
that only publishes (signs) messages, but does not receive (validate) them, so
we should be okay.

### DIFF
--- a/fedmsg/crypto/__init__.py
+++ b/fedmsg/crypto/__init__.py
@@ -155,8 +155,19 @@ log = logging.getLogger(__name__)
 _implementation = None
 _validate_implementations = None
 
+import six
+
+# For python3, we're in trouble due to M2Crypto - it won't be ported.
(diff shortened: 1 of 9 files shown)


pypingou:
Looks fine to me

bkabrda:
:+1: Cool!

PR #344, opened 2015-08-06
Add a new "subjective" api.

ralphbean:
This will allow us to render messages in Fedora Hubs that instead of looking like:

  "ralph submitted a bodhi update for nethack"

can look like:

  "you submitted a bodhi update for nethack"


pypingou:
looks good to me

PR #345, opened 2015-08-12
Reconcile the subjective API with the conglomerate API.

ralphbean:
This should make it so you can pass an optional ``subject`` to the conglomerate
API and have it behave sensibly.

The subjective API allows you get a string like:

- "you submitted an update"

from a message where you would normally get a string like:

- "ralph submitted an update"

**If** you pass it a ``subject=ralph`` argument.

The conglomerate API allows you to group messages together and get single
strings that represent all of them.  So, instead of getting:

- "ralph submitted update X"
- "ralph submitted update Y"
- "ralph submitted update Z"

You get:

- "ralph submitted updates X, Y, and Z"

----

This patch reconciles those two APIs by letting the conglomerate functions
handle a ``subject`` if one is provided, so you can get something like:

- "you submitted updates X, Y, and Z"

This patch changes the conglomerate API and tests and requires a similar patch
provided in fedora-infra/fedmsg_meta_fedora_infrastructure#305.

This will be nice for the fedora-hubs implementation (eventually).

### DIFF
--- a/fedmsg/meta/__init__.py
+++ b/fedmsg/meta/__init__.py
@@ -157,7 +157,7 @@ def with_processor():
     return _wrapper
 
 
-def conglomerate(messages, **config):
+def conglomerate(messages, subject=None, **config):
     """ Return a list of messages with some of them grouped into conglomerate
     messages.  Conglomerate messages represent several other messages.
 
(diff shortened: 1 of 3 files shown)


relrod:
+1 to the idea. I wish there were a more formal way to verify that all of the functions that depend on these changed ones have been changed and non have been missed. But, alas.

ralphbean:
Agreed.  Thanks Ricky!

PR #347, opened 2015-08-22
core: fix gpg signing

mathstuf:
The cert_index variable is not defined in this branch. Instead, allow it
to be defaulted, but fall back to using the hostname as the key.


mathstuf:
The config docs are…less than helpful in places. I ended up digging down to figure out how the gpg key is determined and found this.

The project: https://github.com/mathstuf/eventd-plugin-fedmsg

Issue #121, opened 2013-02-25
Separate out moksha deps


ralphbean:
Not everyone wants the fedmsg-hub mechanics (and the deps!) by default.

People who just want to listen for messages should be able to with fedmsg.tail_messages without pulling in all of twisted.

Kind of like http://www.fedmsg.com/en/latest/consuming/#diy-listening-with-raw-zeromq

yazman:
I'm closing this issue (at long last!) as the above mentioned pull request resolves it.

PR #349, opened 2015-08-26
Isolate Moksha and cryptographic dependencies as extras

yazman:
I found [this semi-ancient issue](https://github.com/fedora-infra/fedmsg/issues/121), and thought it would be a good way to get to know fedmsg better. If it's not relevant anymore, feel free to close out this pull request.

What I've done is split off the non-essential dependencies of fedmsg into what setuptools calls "extras." These are not installed by default but require a command such as ``pip install fedmsg[crypto]``. These allow us to communicate optional dependencies to users. I've created three extras: 'hub', for Moksha and all its dependencies; 'crypto', for the semi-essential cryptographic libraries; and 'all', which is the auto-generated sum of the prior two, plus two other packages which had been previously indicated as non-essential. (It's possible that I misunderstood that role of those two packages, ``daemon`` and ``arrow``, so please correct me if I'm wrong.) As many of the fedmsg commands require the presence of Moksha, I've flagged those entry-points in a manner that setuptools and ``pip`` both understand.

In working on fedmsg, I noticed that ``fedmsg-tail`` doesn't actually work out of the box without M2Crypto. I saw that there were prior plans to add command-line arguments to override the need for it, and I went ahead and implemented them as ``--validate`` and ``--no-validate`` (if a parameter of the form ``--validate=[true|false]`` is preferable, I can easily change that). This is somewhat tangential to the primary feature targeted by this pull request, and I can split it off for separate review if so desired.

Aside from all that, I've also updated the documentation accordingly (I think I've found all places where these changes would be relevant) and bumped up the version number by a minor point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
  - "3.4"
 
 install:
- - python setup.py install
+ - pip install -e .[hub]
  - pip install nose mock sqlalchemy unittest2
 script: nosetests -v
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
  - "3.4"
 
 install:
- - pip install -e .[hub]
+ - pip install -e .[all]
  - pip install nose mock sqlalchemy unittest2
 script: nosetests -v
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
  - "3.4"
 
 install:
- - pip install -e .[all]
+ - pip install -e .[commands,consumers]
  - pip install nose mock sqlalchemy unittest2
 script: nosetests -v
 notifications:
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,10 @@ fedmsg -- Fedora Messaging Client API
 .. split here
 
 Utilities used around Fedora Infrastructure to send and receive messages.
-Please see ``doc/`` or https://fedmsg.readthedocs.org/ for more info.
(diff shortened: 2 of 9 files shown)


yazman:
Over IRC, @ralphbean recommended splitting the package up into extras in a way that mirrors Fedora's RPM structure as much as possible. I agreed and went about implementing the change, which essentially manifested as a 'commands' extra for all the basic commands and a 'consumers' extra for *most* things related to Moksha. In RPM-land, both ``python-fedmsg-commands`` and ``python-fedmsg-consumers``, from which I drew the division of dependencies, require ``python-moksha-hub``.  This seemed really odd, and after digging a little deeper, I found that the only component of ``python-fedmsg-commands`` that was pulling in Moksha was ``fedmsg-collectd``. Given the fact that ``fedmsg-collectd`` is both not a command and a user of the Moksha consumer API, I think it would be most appropriate to move that over to the 'consumers' bucket. *Hopefully*, this can be done with the RPMs as well. If not, then this is a moot point, and I'll push a change here to bring things back in line with the RPM status quo (which is still a plus).

Also, I found that only components of the 'consumers' bucket can potentially require ``deamon`` as a dependency. This is despite the fact that ``python-daemon`` is listed as a dependency of ``python-fedmsg-commands`` but *not* of ``python-fedmsg-consumers``. This is a much more clear-cut issue, as you can see in the code that the ``daemon`` module may only be imported if the attribute ``daemonizable`` equals ``True`` on a ``fedmsg.commands.BaseCommands`` subclass, which only happens in the command classes for ``fedmsg-{gateway,relay,hub,irc}``.

PR #350, opened 2015-09-05
Add a pre_fire_hook callback to fedmsg.publish.

ralphbean:
This can be extended by other libraries to do something before publishing.

### DIFF
--- a/fedmsg/core.py
+++ b/fedmsg/core.py
@@ -175,7 +175,8 @@ class FedMsgContext(object):
 
         return self.publish(topic, msg, modname)
 
-    def publish(self, topic=None, msg=None, modname=None, **kw):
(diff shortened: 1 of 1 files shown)


fedmsg/core.py
```diff
@@ -294,6 +299,9 @@ def publish(self, topic=None, msg=None, modname=None, **kw):
             # Add the seq_id field
             msg = store.add(msg)
 
+        if pre_fire_hook:
```
pypingou: Should we check if it is callable and raise an exception if not?

pypingou:
One question, otherwise :+1: for me

fedmsg/core.py
ralphbean: Well, it will raise an exception when we try to call it.  :)

PR #353, opened 2015-09-16
Add a new msg2agent API call to fedmsg.meta.

ralphbean:
As the docstrings indicate, this is different from msg2usernames.

- msg2usernames returns *all* of the users associated with a message and
  doesn't distinguish their particular relationships with the event.

- This new msg2agent will return *just* the user who is responsible for the
  event.  So, in the case of one user giving karma to another, both users will
  be returned by msg2usernames, but just the one who gave the karma will be
  returned by msg2agent.

Furthermore, this pull request changes the test framework to allow testing None values.

Previously, if the ``expected_{value}`` on a test was set to ``None``, then
that test would be skipped.  However, that didn't allow us to test that a
value returned *really was None* or not.

This change introduces the use of a sentinel value called ``Unspecified``
that we use instead of ``None`` so that we can now set a test to ``None``
over in fedmsg_meta_fedora_infrastructure and find out if the actual value
returned by our code really was None or not.

### DIFF
--- a/fedmsg/meta/__init__.py
+++ b/fedmsg/meta/__init__.py
@@ -278,6 +278,36 @@ def msg2usernames(msg, processor=None, legacy=False, **config):
     return processor.usernames(msg, **config)
 
 
+@with_processor()
+def msg2agent(msg, processor=None, legacy=False, **config):
+    """ Return the single username who is the "agent" for an event.
+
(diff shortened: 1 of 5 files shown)


pypingou:
Looks good to me

ralphbean:
Thanks!

PR #354, opened 2015-10-06
Try three times to get the CRL.

ralphbean:
I've gotten reports from multiple people in recent weeks that their fedmsg
scripts fail and exit when *something happens* and they are unable to pull down
the CRL.  We're not sure why that's failing yet.  It may be a network blip, or
it may be a proxy issue.  In the meantime, we can make fedmsg try a little
harder to get that file before it blows up on the user.


pypingou:
:+1: for me

PR #348, opened 2015-08-22
core: split tail_message

mathstuf:
Split tail_message into:

  - _create_poller
  - _poll
  - _poll_socket
  - _close_subs

This allows the polling to be hooked into another event loop.

---
Not sure if this is the best way to do this.

PR #356, opened 2015-10-07
Stop yielding None values from fedmsg.tail_messages.

### LINKED ISSUE #355: tail_messages yields None
For a long time tail_messages always returned a 4-tuple, but recently it started to return None after receiving invalid message. I think it's not intended

ralphbean:
This fixes #355.

The bug got introduced in #348.


pypingou:
:+1: for me

PR #360, opened 2016-03-03
use setsockopt_string when setting a string as the option

AdamWill:
setsockopt() accepts, strictly, an int or a bytestring as the
option value. Here we want to use a string as the value, so we
should use setsockopt_string(). I checked all other uses and
they look to be using ints as the option values, so they're OK.

Without this fix, reading the first message from tail_messages
will blow up in Python 3.

See https://pyzmq.readthedocs.org/en/latest/api/zmq.html#zmq.Socket.setsockopt_string .


ralphbean:
Perfect.

(Alternatively we could ``.encode('utf-8')`` the topic before passing to ``setsockopt``, but this works just as well.)

PR #361, opened 2016-03-03
ensure we pass a str not a bytes to loads()

AdamWill:
another Python 3 fix, again, without this, tail_messages()
blows up pretty fast. In Python 3, message here is a bytes,
it's not a str. loads specifically needs a str.

Note: there are several other suspicious-looking `loads()` calls throughout fedmsg, but it's very very late and I do not feel like investigating them all...

### DIFF
--- a/fedmsg/core.py
+++ b/fedmsg/core.py
@@ -405,7 +405,7 @@ class FedMsgContext(object):
         validate = self.c.get('validate_signatures', False)
 
         _topic, message = sock.recv_multipart()
-        msg = fedmsg.encoding.loads(message)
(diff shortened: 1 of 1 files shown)


ralphbean:
I think the more suspicious clue to follow is any sort of call to ``sock.whatever(...)``.  That will be returning byte strings which we should decode before doing anything else.

In other places where loads is called, we should hopefully be able to assume that unicode strings are already in play.

ralphbean:
This is good, but I'm going to close it in favor of #363.  Let's continue discussion there.

PR #364, opened 2016-03-03
instead of setsockopt_string, encode topic and use setsockopt

AdamWill:
acc7260 was bad - it blows up in Python 2 because there the
value is not a unicode, and setsockopt_string only accepts
unicodes. Making the value a unicode on Python 2 without
crashing in Python 3 attempting to decode a str is annoying,
so instead let's encode the topic and use setsockopt.

This will blow up in Python 3 if the topic is already encoded,
but I don't think that should happen.


ralphbean:
:+1: .. let's get a new release out :steam_locomotive:

PR #366, opened 2016-03-17
Add long_form metadata to the constituents of a grouped event.

ralphbean:
Mostly just because I want to use it in fedora-hubs.

We'll use this to embed summaries about blog posts and mailing list threads in your feed!


pypingou:
Easy enough :+1:

ralphbean:
Thanks @pypingou!

PR #367, opened 2016-03-17
Add the possibility to associate a pygments lexer with a message.

ralphbean:
This is for fedora-hubs.

We add an implementation in fedora-infra/fedmsg_meta_fedora_infrastructure#367
which returns a DiffLexer or a MarkdownLexer for certain kinds of messages.
This will let us render the long-form of those messages nicely in your feed.

### DIFF
--- a/fedmsg/meta/base.py
+++ b/fedmsg/meta/base.py
@@ -168,6 +168,13 @@ class BaseProcessor(object):
         """ Return some paragraphs of text about a message. """
         return ""
 
+    def lexer(self, msg, **config):
+        """ Return a pygments lexer that can be applied to the long_form.
+
+        Returns None if no lexer is associated.
+        """
+        return None
+
     def link(self, msg, **config):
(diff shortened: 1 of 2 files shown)


fedmsg/meta/base.py
```diff
@@ -168,6 +168,13 @@ def long_form(self, msg, **config):
         """ Return some paragraphs of text about a message. """
         return ""
 
+    def lexer(self, msg, **config):
+        """ Return a pygments lexer that can be applied to the long_form.
+
+        Returns None if no lexer is associated.
+        """
+        return None
```
pypingou: Should we return a default one?

pypingou:
One question but :+1:

fedmsg/meta/base.py
ralphbean: I had that originally, but I couldn't find a sensible default.  Using None works as long as the caller checks for it.

PR #368, opened 2016-03-19
Update ircbot.py

glensc:
add [link](https://github.com/fedora-infra/fedmsg/blob/0.16.4/fedmsg/consumers/ircbot.py#L48-L65) to available irc bot colors


ralphbean:
Looks good.  Thanks again @glensc!

PR #369, opened 2016-03-19
irc: default color configurable

glensc:
i use light terminal, so text in 'light grey' color is unreadable


ralphbean:
Looks good to me.  Thanks!

PR #374, opened 2016-07-08
Cascade IRC connections.

ralphbean:
This makes it so that, if we are joining 12 channels as 12 bots, we won't try
to do them all at the same time.  We'll connect to one, and then once that is
started we'll wait 1 second and then start the second connection.  The start of
that one will wait 1 second and trigger the third, etc...


puiterwijk:
Looks good to me. :+1:

Issue #376, opened 2016-07-21
Drop the support for Python 2.6


tvieira:
I think we should drop the support for python 2.6. Many dependencies (i.e. Twisted, which is a dependency for moksha.hub) already dropped the support for 2.6. To go forward we would have to lock versions which increases the complexity of maintenance.

ncoghlan:
I just ran into this, with a docs PR failing due to Python 2.6 still being in the test matrix, but Twisted not being capped to 15.4 on Python 2.6.

Theoretically, the restriction could be limited to just Python 2.6 by added "twisted < 15.5; python_version < '2.7'" to the requirements, but then you run into the problem of needing a sufficiently recent pip that understands requirements markers.

If there's still a use case where fedmsg needs to run in the system Python on RHEL 6, then the simplest option might be to add the "twisted < 15.5" restriction universally until Fedora Infrastructure only needs to worry about RHEL 7 and Fedora.

ashcrow:
If it does need to run on RHEL 6 there is always [Software Collections](https://www.softwarecollections.org/en/scls/rhscl/python27/)

PR #379, opened 2016-10-08
Fix and unite local and hosted testing

ashcrow:
I noticed attempting to run ``tox -vv`` and ``python setup.py nosetests`` both failed locally unless a number of packages were manually installed first. I also saw there were a few missing packages in the list of requirements. This change adds those testing requirements in and uses the same list for ``tox``, ``nosetests``, and ``travis``.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
- - "2.6"
  - "2.7"
  - "3.3"
  - "3.4"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ python:
 
 install:
  - pip install -e .[commands,consumers]
- - pip install nose mock sqlalchemy unittest2
-script: nosetests -v
+ - pip install nose
+script: python setup.py nosetests
 notifications:
     email: false
     irc:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ python:
  - "2.7"
  - "3.3"
  - "3.4"
+ - "3.5"
 
 install:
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -1,14 +1,14 @@
 language: python
 python:
- - "2.6"
```
pypingou: Why dropping that one?

ashcrow: @pypingou Twisted dropped support for 2.6. Since fedmsg uses Twisted the builds will always fail. See https://github.com/fedora-infra/fedmsg/issues/376.

.travis.yml
```diff
@@ -7,8 +7,8 @@ python:
 
 install:
  - pip install -e .[commands,consumers]
- - pip install nose mock sqlalchemy unittest2
```
ashcrow: The feature used in``unittest2`` is no longer needed for Python 2.7+.

jeremycline:
I have one request before this is merged: I'd like to keep the Python 2.6 test runner going because there are still some EL6 hosts around. They're running RPM installs with Twisted 12.2, so as long as we don't introduce Python 2.6 incompatible code, the fact that Twisted has dropped 2.6 support shouldn't be a problem.

We just need to make sure that pip installs twisted < 15.4 (it sounds like) for 2.6 in Travis.

ashcrow:
@jeremycline if you don't mind doing the follow up change I'd appreciate it.

ashcrow:
@jeremycline, @bowlofeggs, @pypingou thank you all for your reviews!

jeremycline:
@ashcrow Thanks so much! :shipit:

PR #380, opened 2016-10-09
Get fedmsg-hub working on STOMP.

ralphbean:
The goal here is to be able to take the litany of apps we've written (FMN, fedmsg-irc, datanommer, fedimg, etc..) and be able to run them in other messaging environments (in particular, the RH unified message bus).

I intend to support:

- The hub/consumer approach.  If you have a fedmsg consumer, it should be able
  to listen to messages over other protocols now.
- Publishing with `fedmsg.publish`.  This will only work *from* a fedmsg
  consumer.  It will figure out the moksha context and ask *it* to publish on
  fedmsg's behalf, thereby publishing over STOMP or AMQP, depending on
  configuration.

I explicitly am *not* supporting:

- The naive listening approach, i.e. `fedmsg.tail_messages()`.  It would be a
  ton of work and its not worth it atm.  If we find a *really* good reason to
  implement support for it, we can always revisit this later.

### DIFF
--- a/doc/compatibility.rst
+++ b/doc/compatibility.rst
@@ -33,7 +33,10 @@ In order to get a fedmsg-hub instance working with STOMP, you should add the
 following configuration to `/etc/fedmsg.d/base.py`::
 
     # We almost always want the fedmsg-hub to be sending messages with zmq as
-    # opposed to amqp or stomp.
+    # opposed to amqp or stomp.  You can send with only *one* of the messaging
+    # backends: zeromq or amqp or stomp.  You cannot send with two or more at
+    # the same time.  Here, zmq is either enabled, or it is not.  If it is not,
+    # see the options below for how to configure stomp or amqp.
     #zmq_enabled=True,
 
     # On the other hand, if you wanted to use STOMP *instead* of zeromq, you
--- a/doc/compatibility.rst
+++ b/doc/compatibility.rst
@@ -1,0 +1,52 @@
+Compatibility with Other Messaging Technologies
+===============================================
+
+fedmsg was originally built to specifically support zeromq as its messaging backend.
+
+However, we also originally built it on top of `moksha <http://moksha.ws>`_
+which is a framework that already supports not only zeromq, but also STOMP and
+AMQP. In 2016, we added the ability for fedmsg to send and receive messages
+over those protocols (and any others that moksha ends up supporting).
+
+This lets you take advantage of the litany of apps written around the fedmsg
+ecosystem (FMN, fedmsg-irc, datanommer, fedimg, etc..) and run them in other
+messaging environments (STOMP, AMQP, ...).
+
+We intentionally support:
+
+- *The hub/consumer approach*. If you have a fedmsg consumer, it should be able
+  to listen to messages over other protocols now.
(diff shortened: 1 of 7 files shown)


ralphbean:
FYI, the code here is *probably* ready to go, but I'm waiting for a dev broker to become available to do some integration testing.  Let's wait to merge this until I can confirm things work at that level.

pypingou:
Code wise it looks good to me, maybe we could just document somewhere how the configuration file should look like then (since it needs the moksha hub to be correctly configured).

Let's wait for your tests and then :+1: to merge for me.

ralphbean:
OK, this works now!  It depends on mokshaproject/moksha#36 too..

ralphbean:
And, docs available now too.  :+1: to merge from me.

Any concerns?

fedmsg/meta/base.py
```diff
@@ -154,7 +154,9 @@ def handle_msg(self, msg, **config):
             return match.groups()[-1] or ""
 
     def title(self, msg, **config):
-        return '.'.join(msg['topic'].split('.')[3:])
+        if msg['topic'][0].isalpha():
```
pypingou: This checks that the first letter of the topic is alphabetic, right? What about checking that there are at least 3 `.` in there?

Also, thoughts on using ``msg['topic'].split('.', 3)[-1]`` below?

pypingou:
Couple of questions but still looking good :)

fedmsg/meta/base.py
ralphbean: FYI, the topic structure for stomp looks like ``/topic/Consumer....``.

Could I check instead if `if msg['topic'].startswith('/topic/')` ?

doc/compatibility.rst
```diff
@@ -0,0 +1,52 @@
...
+  reason to implement support for it, we can always revisit this later.
+
+Configuration
+-------------
+
+In order to get a fedmsg-hub instance working with STOMP, you should add the
+following configuration to `/etc/fedmsg.d/base.py`::
+
+    # We almost always want the fedmsg-hub to be sending messages with zmq as
+    # opposed to amqp or stomp.
```
ralphbean: Yup.

fedmsg/meta/base.py
ralphbean: Added in 9159488

doc/compatibility.rst
ralphbean: Fixed in 5df1e48.

fedmsg/meta/base.py
pypingou: Cool, what about the `msg['topic'].split('.', 3)[-1]`? Not that it matters more just an idea

pypingou:
:+1: for me

fedmsg/meta/base.py
ralphbean: It makes my head spin trying to imagine all the different combinations of dots..  :dizzy_face:

ralphbean:
Thanks @pypingou!

fedmsg/meta/base.py
pypingou: fair :)

ralphbean:
Can you merge it when you have a moment?

pypingou:
Sure thing

PR #381, opened 2016-10-27
Raise the resource limit on open files for fedmsg-hub

jeremycline:
This is a short-term solution to systems running into the maximum open
file descriptor limit.

### DIFF
--- a/fedmsg/commands/hub.py
+++ b/fedmsg/commands/hub.py
@@ -17,9 +17,18 @@
 #
 # Authors:  Ralph Bean <rbean@redhat.com>
 #
-import fedmsg
+from gettext import gettext as _
+import resource
+
 from fedmsg.utils import load_class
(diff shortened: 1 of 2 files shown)


ralphbean:
:+1: from me!

fedmsg/commands/hub.py
```diff
@@ -17,11 +17,20 @@
 #
 # Authors:  Ralph Bean <rbean@redhat.com>
 #
-import fedmsg
+from gettext import gettext as _
```
pypingou: do we already depend on gettext?

jeremycline: [gettext](https://docs.python.org/2/library/gettext.html) is in the standard library. I don't know exactly when it was introduced, but it's definitely 2.6+ so we're good.

pypingou:
Cool :+1: for me as well then :)

PR #382, opened 2016-10-27
Turn testing Python 2.6 in Travis on

jeremycline:
Twisted has dropped Python 2.6 support, but fedmsg still needs to be
Python 2.6 compatible. Since moksha.hub, the dependency that pulls in
Twisted, doesn't require a specific version of Twisted, we can just
install it before we install the requirements if the environment is
Python 2.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
- - "2.6"
  - "2.7"
  - "3.3"
  - "3.4"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ python:
 
 install:
  - pip install -e .[commands,consumers]
- - pip install nose mock sqlalchemy unittest2
-script: nosetests -v
+ - pip install nose
+script: python setup.py nosetests
 notifications:
     email: false
     irc:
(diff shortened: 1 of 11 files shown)


jeremycline:
This pulls in all the commits for #379 - we can either merge this and close #379, or merge #379 and rebase this. I do not mind either way.

.travis.yml
```diff
@@ -4,11 +4,14 @@ python:
  - "2.7"
  - "3.3"
  - "3.4"
+ - "3.5"
 
+# Twisted dropped Python 2.6 support, so install a compatible version on 2.6
 install:
+ - if [[ $TRAVIS_PYTHON_VERSION == 2.6 ]]; then pip install Twisted==12.2; fi
  - pip install -e .[commands,consumers]
- - pip install nose mock sqlalchemy unittest2
```
pypingou: isn't unittest2 required in 2.6?

jeremycline: It is, and it's placed in the tests_require list in setup.py

jeremycline:
I added a ``test-requirements.txt`` file with all the dependencies so we can pip install them. Simply doing ``python setup.py nosetests`` to install the test requirements was causing installations to fail on 2.6 (for reasons I did not dig into too deeply).

pypingou:
:+1: for me

PR #386, opened 2016-11-09
Add SSL support to irc bot

asdil12:


setup.py
```diff
@@ -87,6 +87,8 @@
         'daemon',      # not *necessarily* required
         'pygments',
         'psutil',
+        'pyopenssl',
```
ralphbean: Should these new deps be optional?

asdil12:
@ralphbean like this?

ralphbean:
@asdil12, yeah, that looks good.  Thank you :)

PR #387, opened 2016-11-10
Document turning off validation for other busses.

ralphbean:
There's usually no point in cryptographically validating messages from other
busses.  They likely won't bear message certificates and signatures. Other
busses usually use TLS on the connection itself for authentication and
authorization.


ralphbean:
@mprahl, can you review?

PR #388, opened 2016-11-10
Return earlier when validate_signatures is turned off

mprahl:
When connecting to a bus without messages being signed, an exception would be raised with the message "Topic envelope mismatch". If signing is turned off, none of the message contents should be examined in the `validate` function.


ralphbean:
Looks good here.  :)

PR #390, opened 2016-11-10
Drop old testcases against old python-six.

ralphbean:
It used to be 1.7 on RHEL7, but they upgraded and we only have to worry about 1.9 now.


pypingou:
Looks fine to me

ralphbean:
Thanks @pypingou!

PR #393, opened 2016-12-01
Only check for stomp messages after we decoded any ZMQMessage

puiterwijk:
Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>


ralphbean:
Hm, this is reverting more than it needs to.  The stuff from #388 should be ok.

ralphbean:
After reading through a second time, this looks good now.

Issue #304, opened 2014-12-01
python sample does not work like suggested in docs


glensc:
Python sample from http://fedmsg.readthedocs.org/en/latest/#publishing-messages-with-python
does not seem to work without extra configuration:

```
$ python test.py 
/usr/share/python2.7/site-packages/fedmsg/core.py:153: UserWarning: fedmsg is not configured to send any messages for name '__main__.rotten-fruit'
Traceback (most recent call last):
  File "test.py", line 4, in <module>
    'test': "Hello World",
  File "/usr/share/python2.7/site-packages/fedmsg/__init__.py", line 71, in _wrapper
  File "/usr/share/python2.7/site-packages/fedmsg/__init__.py", line 86, in publish
  File "/usr/share/python2.7/site-packages/fedmsg/core.py", line 296, in publish
AttributeError: 'FedMsgContext' object has no attribute 'publisher'

$ cat test.py 
import fedmsg

fedmsg.publish(topic='testing', modname='test', msg={
    'test': "Hello World",
})
$ hostname
rotten-fruit
```

the solution is to fill `/etc/fedmsg.d/endpoints.py`, which is not mentioned in that doc

```
21:07:23  threebean> you need a '__main__.rotten-fruit' entry in /etc/fedmsg.d/endpoints.py
21:15:46  threebean> glen: something like this -> http://fpaste.org/155568/14174613/
```

``` python
config = dict(
    # This is a dict of possible addresses from which fedmsg can send
    # messages.  fedmsg.init(...) requires that a 'name' argument be passed
    # to it which corresponds with one of the keys in this dict.
    endpoints={
        "__main__.rotten-fruit": [
            "tcp://127.0.0.1:4321",
        ]
    },
)
```

and it still does not print output to fedmsg-tail, will be updating this ticket when i get more information.

Velmont:
I was really bitten in the ass about this too.  Maybe it would be good to mention something about this not really working?

frenzymadness:
IMHO this issue can be closed.

glensc:
if the fix is changing some timeout, then it's not resolved imho.

```
21:40:20 glen> post_init_sleep=0.5,
21:40:24 glen> is in default base.py
21:40:52 glen> changing to 1 seems better indeed
```

PR #394, opened 2016-12-13
Update publishing sample code

sijis:
This is an attempt to ensure the example code in the publishing section works.

See #304 for detail discussion on the issue.


doc/index.rst
```diff
@@ -49,10 +49,12 @@ See :doc:`development` on setting up your environment and workflow.
 .. code-block:: python
 
    import fedmsg
-   fedmsg.publish(topic='testing', modname='test', msg={
+   fedmsg.publish(name='mybus', topic='testing', modname='test', msg={
```
pypingou: The name is optional, isn't it?

sijis: Yes it is, however, adding that parameter makes the example code easier to have work (and explain).
Otherwise, we have to explain that the default name is `__main__.$(hostname -s)`, and that should be added as the entry in `endpoints.py` instead of something like `mybus`.

pypingou: Should we document both then?

sijis: We could. It is a good idea actually.

So do you suggest I leave the original sample code and just note the `endpoint.py` entry based if the `name` parameter in `publish()` is used or not?

pypingou: Sounds good, or provide both example and explain how the ``endpoint.py`` should look in both cases

sijis:
@pypingou Added some additional changes. Let me know what you think.

sijis:
That's better said. I'll update the content accordingly.

doc/index.rst
```diff
@@ -46,13 +46,31 @@ Publishing Messages with Python
...
+
 .. code-block:: python
 
    import fedmsg
    fedmsg.publish(topic='testing', modname='test', msg={
        'test': "Hello World",
    })
 
+.. note:: The ``endpoints.py`` file should have an entry as ``"__main__.myhost": [ "tcp://127.0.0.1:4321", ]``. The
+   value of ``__main__.myhost`` is autogenerated by default, and ``myhost`` is presented by the equivalent output
```
sijis: Updated the content.

pypingou:
👍 for me, looking good! :)

sijis:
@glensc @odinho Do you think these doc changes clear up the confusion raised in issue #304?

odinho:
Yes, that looks very nice to me :)

glensc:
i will omit my answer in this case

pypingou:
> i will omit my answer in this case

I have no idea how to interpret this, do you not like the change and prefer to say nothing than saying something bad? You don't really care about the change so have nothing to say? You do like the change but you would prefer something else?

It would be really appreciated if you could extend a little on your opinion. This would help us improve the documentation or consider this as good to go and merge :)

Thanks in advance!

glensc:
merge it. keep up the good work. the bugreport is 2+ years old. and i don't use fedmsg, it takes too much effort for me to dig back in.

pypingou:
Alright let's merge then :)

Thanks for your inputs everyone and thanks for the patch @sijis !

PR #395, opened 2017-01-18
Fedmsg-relay shouldn't run producers.

ralphbean:
If another package happens to be installed in the system, running
`fedmsg-relay` shouldn't inadvertently start that thing.


pypingou:
Fair, 👍

ralphbean:
Thanks! :bear:

PR #396, opened 2017-01-23
New hmac backend to sign and validate messages

pypingou:
This backend should make it easier for people to use fedmsg for smaller
deployment than the one used by Debian or the Fedora infrastructure and
where it is easier to keep the secret used... secret.
It should also make it easier to run fedmsg for development while still
having signed messages.

### DIFF
--- a/fedmsg/core.py
+++ b/fedmsg/core.py
@@ -288,7 +288,10 @@ class FedMsgContext(object):
 
         # Find my message-signing cert if I need one.
         if self.c.get('sign_messages', False):
-            if not self.c.get("crypto_backend") == "gpg":
+            if self.c.get("crypto_backend") == "gpg":
+                if 'gpg_signing_key' not in self.c:
+                    self.c['gpg_signing_key'] = self.c['gpg_keys'][self.hostname]
+            else:
                 if 'cert_prefix' in self.c:
                     cert_index = "%s.%s" % (self.c['cert_prefix'],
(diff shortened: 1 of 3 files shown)


pypingou:
I will be working on the tests but wanted to offer the core logic for review as early as possible

pypingou:
> Also, note that when you enable multiple validation backends, fedmsg will return success if any of the enabled ones return success.
This means that if you have x509 and hmac enabled in the same system, every message topic can be sent by any sender, and since hmac doesn't follow routing policies, it would be a particular weak point in such a set up.

I am not sure you can do this since ``crypto_backend`` in the configuration is a string, not a list.

puiterwijk:
> I am not sure you can do this since crypto_backend in the configuration is a string, not a list.

```
for mod in config.get('crypto_validate_backends', []):
    _validate_implementations.append(fedhmac)
```

:-)

pypingou:
Indeed for message validation it supports validating different type of signatures.

jeremycline:
Longer-term I'd prefer us to get out of the message authentication game and use [upstream's solution](http://api.zeromq.org/4-2:zmq-curve) with authorization built on top of that (and, of course, easy configuration). This, of course, means having a higher ZMQ version requirement, but it also means not dealing with cryptography much ourselves.

Given that, I don't think we should add to the message authentication mechanisms we already have. Are you okay with me closing this PR?

PR #398, opened 2017-02-06
Allow fedmsg-tail to be daemonizable

pabelanger:
Fedmsg-tail is a great tool for testing, however it currently required
2 terminals (or screen) to be used properly.  Fedmsg already support
other commands to be run as daemons, so lets add that support to
fedmsg-tail.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>

### DIFF
--- a/fedmsg/commands/tail.py
+++ b/fedmsg/commands/tail.py
@@ -38,6 +38,7 @@ class TailCommand(BaseCommand):
     """ Watch all endpoints on the bus and print each message to stdout. """
 
     name = "fedmsg-tail"
+    daemonizable = True
     extra_args = [
         (['--topic'], {
             'dest': 'topic',


pypingou:
Well, fedmsg-tail is meant to be used to show the traffic of a bus on the screen, so I'm not quite sure it makes sense to have it be a daemon.
Where would the output be printed in daemon mode?

pabelanger:
/var/log/fedmsg/fedmsg-tail.log

This is helpful so I don't have to write my own logic for a screen session. In my example below, we are just using fedmsg-tail as part of an integration test, to validate fedmsg.publish() works correctly.

https://review.openstack.org/#/c/426861/17/tools/nodepool-integration-setup.sh

pypingou:
Hm looks like this impact some of the tests

jeremycline:
Hi @pabelanger, what I tend to do is run fedmsg-tail as a systemd-controlled service. Would that work for you?

More generally I'm wondering if we want to drop the daemonization support in fedmsg and just have users use systemd.

ralphbean:
> More generally I'm wondering if we want to drop the daemonization support in fedmsg and just have users use systemd.

:+1: to this.  :)

jeremycline:
Since we've decided to deprecate the daemon feature entirely, I'm going to go ahead and close this PR. Thanks, though!

PR #402, opened 2017-03-14
Remove artifical keys from datagrepper backlog to pass validation.

ralphbean:
Today, when fedmsg-hub starts up, it tries to read in its backlog of
messages from datagrepper in case it missed anything.  It successfully
gets the list, but then every single one of the messages fails
validation.

There's a gist here that demonstrates what needs to happen to make this
work:  https://gist.github.com/ralphbean/b532ad3bc4a8333fa03d4e4e48451885

There are two changes required.  This one here, in fedmsg, and another in
fedora-infra/datanommer#87.


pypingou:
👍 for me

PR #406, opened 2017-04-05
Remove env validation.

ralphbean:
We have a case where we want the environment to be something other than dev,
stg, or prod.  I figure there's not really a reason to enforce any convention
in the code like this.


ralphbean:
Failing tests should be fixed by #405.

ralphbean:
Rebased on top of #405.

pypingou:
Looks fine to me :)

ralphbean:
Thanks @pypingou!

PR #408, opened 2017-05-02
Add explicit types to the publish API

jeremycline:
I had some questions this morning about what types I could (or rather,
what I _should_) give to fedmsg. The documentation wasn't clear, so I
have made it clear.

I used the parameter documentation format for sphinx and shuffled things
around a bit as part of this.


pypingou:
Looks fine to me

PR #409, opened 2017-05-15
Add a signing relay

jeremycline:
This adds a new relay type and command that signs incoming messages
prior to relaying them.


pypingou:
Looks fine to me

PR #414, opened 2017-06-04
minor refactors/pep8tification

pravinkc:
I was going thru the codebase and added some refactors/comments etc.


fedmsg/core.py
```diff
@@ -163,6 +164,7 @@ def __init__(self, **config):
             pass
 
         # Cleanup.  See https://bit.ly/SaGeOr for discussion.
+        # arg signature - weakref.ref(object [, callback])
```
bowlofeggs, line 166: Can you explain this new line?

fedmsg/core.py
```diff
@@ -369,7 +372,7 @@ def _create_poller(self, topic="", passive=False, **kw):
 
         # TODO -- the 'passive' here and the 'active' are ambiguous.  They
         # don't actually mean the same thing.  This should be resolved.
-        method = passive and 'bind' or 'connect'
+        method = (passive and 'bind') or 'connect'
```
bowlofeggs, line 374: This change doesn't seem necessary, though I also don't think it is harmful. The order of operations already works this way.

fedmsg/core.py
```diff
@@ -463,7 +467,7 @@ def _run_socket(self, sock, name, ep, watched_names=None):
                     else:
                         raise ValidationError(msg)
             else:
-                return name, ep, _topic, msg
+                return (name, ep, _topic, msg)
```
bowlofeggs, line 468: This change doesn't seem necessary - it's functionally equivalent.

bowlofeggs (approved):
Not being at all familiar with the codebase, it seems reasonable to me.

fedmsg/core.py
pravinkc: I just put the argument signature here, in case someone doesn't know about it (l had to look it up for example)

fedmsg/core.py
pravinkc: I think this makes it more readable. Also, from the zen of python "Explicit is better than implicit"

fedmsg/core.py
pravinkc: equivalent true, it's just more explicit this way i think?
i can remove it (and other things you mention :top: if required :+1:)

fedmsg/core.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, **config):
         self.log = logging.getLogger("fedmsg")
 
         self.c = config
-        self.hostname = socket.gethostname().split('.', 1)[0]
+        self.hostname = socket.gethostname().split('.', maxsplit=1)[0]
```
jeremycline: ``maxsplit`` is a positional argument in Python 2.7, which we unfortunately still have to support.

jeremycline (changes requested):

fedmsg/core.py
pravinkc: great catch, fixed!

bowlofeggs:
On Wed, 2017-06-07 at 00:01 -0700, pravinkc wrote:
> equivalent true, it's just more explicit this way i think?
> i can remove it (and other things you mention 🔝 if required 👍)

It's certainly not required - I'm not opposed to the changes. Explicit
is good ☺

jeremycline:
CI still seems to be unhappy:
```
FAIL: send_message is deprecated
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/fedora-infra/fedmsg/fedmsg/tests/test_core.py", line 29, in test_send_message
    assert str(w[0].message) == ".send_message is deprecated."
AssertionError:
    <module 'warnings' from '/home/travis/virtualenv/python2.6.9/lib/python2.6/warnings.pyc'>.simplefilter("always")
    self.ctx.send_message(topic='org.fedoraproject.prod.compose.rawhide.complete', msg="{'arch'': 's390', 'branch': 'rawhide', 'log': 'done'}")
    assert len([<warnings.WarningMessage object at 0x254d650>]) == 1
>>  assert str([<warnings.WarningMessage object at 0x254d650>][0].message) == ".send_message is deprecated."
    assert self.ctx.publish.called
    topic, msg, modname = self.ctx.publish.call_args[0]
```

Can you look into that?

Note that the issue about importing OrderedDict on Python 2.6 has been fixed in https://github.com/mokshaproject/moksha/pull/41 and just needs a Moksha release before it's good to go.

pravinkc:
@jeremycline I have fixed the tests and went on to add flake8 to the test suite. 
only py26 fails because now because of moksha as you mentioned :top:

.travis.yml
```diff
@@ -1,17 +1,24 @@
...
  - "3.4"
  - "3.5"
 
-# Twisted dropped Python 2.6 support, so install a compatible version on 2.6
 install:
+ # Twisted dropped Python 2.6 support, so install a compatible version on 2.6
  - pip install --upgrade pip setuptools
  - pip install -r test-requirements.txt
  - pip install -e .[commands,consumers]
+ - pip install codecov
```
jeremycline: I've actually got this going in a separate PR (#417), but I like where your head is at :smile:

fedmsg/core.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, **config):
             module_name = guess_calling_module(default="fedmsg")
             config["name"] = module_name + '.' + self.hostname
 
-            if any(map(config["name"].startswith, ['fedmsg'])):
```
jeremycline, line 78: Whoa

fedmsg/core.py
```diff
@@ -139,7 +139,8 @@ def __init__(self, **config):
                     # If we fail to bind or connect, there's probably another
                     # process already using that endpoint port.  Try the next
                     # one.
-                    pass
+                    self.log.debug(
```
jeremycline: Technically this isn't a PEP-8 change or minor refactor, so although I'd definitely accept a change like this, I'd prefer it to be in a separate PR

fedmsg/core.py
```diff
@@ -379,6 +382,7 @@ def _create_poller(self, topic="", passive=False, **kw):
             # it appears in the endpoints list due to a hack where it gets
             # added in __init__ above.
             if _name == 'relay_inbound':
+                self.log.debug("ignored relay_inbound from config.endpoints")
```
jeremycline: Same deal as above. I'd definitely accept a PR that improves the logging, but it's not really related to the other stuff in this PR.

fedmsg/core.py
jeremycline: Personally I see them as equivalent in explicitness, but it's not surprising to run across this and it doesn't hurt anything so I don't mind one way or the other.

As a side node, something I think _is_ more explicit and something I quite like in this sort of scenario is a namedtuple. However I think that's probably better as a separate PR.

fedmsg/tests/fedmsg-test-config.py
```diff
@@ -26,11 +26,12 @@
 here = os.getcwd()
 hostname = socket.gethostname().split('.', 1)[0]
 
-ssl_enabled_for_tests = True
+import imp
```
jeremycline: ``imp`` is actually a deprecated module since 3.4 so we shouldn't use it (even though we're not quite Python 3 compatible yet, but soon!). In addition, I don't know of a good reason to _not_ check for a module's presence by attempting to import it.

fedmsg/tests/test_crypto_switching.py
```diff
@@ -40,10 +41,10 @@ def skip_if_missing_x509_libs(f):
     def _wrapper(self, *args, **kw):
         try:
             if six.PY3:
-                import cryptography
+                imp.find_module('cryptography')
```
jeremycline: Same note as above about the ``imp`` module.

To get it to pass flake8 you can do ``import cryptography  # noqa: E402`` which I think is acceptable in this situation.

fedmsg/tests/test_crypto_x509.py
```diff
@@ -40,8 +41,8 @@
 def skip_if_missing_libs(f):
     def _wrapper(self, *args, **kw):
         try:
-            import M2Crypto
-            import m2ext
+            imp.find_module('M2Crypto')
+            imp.find_module('m2ext')
```
jeremycline: Same note about ``imp`` as above :smile:

fedmsg/tests/test_meta.py
```diff
@@ -74,7 +75,7 @@ def skip_if_fedmsg_meta_FI_is_present(f):
     """
     def _wrapper(self, *args, **kw):
         try:
-            import fedmsg_meta_fedora_infrastructure
+            imp.find_module('fedmsg_meta_fedora_infrastructure')
```
jeremycline: Same note as above about ``imp``

scripts/fedmsg-map.py
```diff
@@ -61,7 +63,7 @@ def scan_all():
     info()
 
     if 'verbose' in sys.argv:
-        import pprint;
+        import pprint
```
jeremycline: Could you move this import to the top of the file?

tox.ini
```diff
@@ -1,7 +1,11 @@
 [tox]
-envlist = py{27}-six{19,Latest}
+envlist = py26,py27-six19,py27-sixLatest,py33,py34,py35,flake8
```
jeremycline: I'd prefer changes to the tox file to occur in a separate PR (and I've got a lot of this in #417 already 😃 )

test-requirements.txt
```diff
@@ -9,3 +9,6 @@ ordereddict; python_version <= '2.6'
 logutils; python_version <= '2.6'
 unittest2; python_version <= '2.6'
 Twisted==12.2; python_version <= '2.6'
+# flake8 3 dropped support for py{26,33}
```
jeremycline: I'm fine only running flake8 on 2.7 since it's for developers and we should all have at least 2.7

jeremycline (changes requested):
Thanks so much for this, it's really great.

I have a few notes inline, but I'm excited for this!

pravinkc:
I have made the requested changes, @jeremycline :+1:

jeremycline (approved):

jeremycline:
Great, thanks so much!

:shipit:

PR #416, opened 2017-06-07
Add a fedmsg-check command

### LINKED ISSUE #415: Add a command to monitor consumers
Moksha already has the (undocumented) ability to set up a publishing socket that appears to announce the running consumers and producers on ``moksha.monitoring.socket``. I would like to have a command that, when provided with a consumer name, subscribes to that socket, gets the announcement, and exits 0 if the consumer is running and 1 if the consumer is not running. This is motivated by wanting to define a health check for a fedmsg consumer in OpenShift.

jeremycline:
This command makes use of the moksha monitor socket to determine the
health of consumers and producers.

fixes #415

### DIFF
--- a/fedmsg/commands/check.py
+++ b/fedmsg/commands/check.py
@@ -1,0 +1,97 @@
+# This file is part of fedmsg.
+# Copyright (C) 2017 Red Hat, Inc.
+#
+# fedmsg is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# fedmsg is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with fedmsg; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+# Authors: Jeremy Cline <jeremy@jcline.org>
+"""
+The ``fedmsg-check`` command can be used to see what fedmsg consumers and
+producers are currently active. It requires that the ``moksha.monitoring.socket``
+is defined in the fedmsg configuration.
+"""
+from gettext import gettext as _
+import json
+
+import click
+import zmq
+
+from fedmsg.config import load_config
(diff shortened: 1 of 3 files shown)


fedmsg/commands/check.py
```diff
@@ -0,0 +1,101 @@
...
+    """This command is used to check the status of consumers and producers.
+
+    If no consumers and producers are provided, the status of all consumers and producers
+    is printed.
+
+    Args:
+        timeout (int): The length of time in seconds to wait for a ZMQ message
+            on the monitor socket.
+        consumer (tuple): A list of consumer names to check the status of.
+        producer (tuple): A list of producer names to check the status of.
```
bowlofeggs: I recommend against documenting the args like this on click commands - if the user types ```--help``` on this command, this entire docblock will be displayed and won't be particularly helpful or useful to the user. It will also give duplicate info since the help will also already tell them about the flags.

See https://github.com/fedora-infra/bodhi/issues/1457

fedmsg/commands/check.py
```diff
@@ -0,0 +1,101 @@
...
+                    missing = True
+                else:
+                    if active[messager]['initialized'] is not True:
+                        click.echo('"{}" is not initialized!'.format(messager), err=True)
+                        uninitialized = True
+
+        if missing:
+            raise click.ClickException('Some consumers and/or producers are missing!')
+        elif uninitialized:
+            raise click.ClickException('Some consumers and/or producers are uninitialized!')
```
bowlofeggs, line 93: Might you want this to be an ```if``` instead of an ```elif```? I.e., if some are missing and others are uninitialized, it'd be nice to report both to the user instead of one or the other. Of course, that breaks the final ```else```, so you probably need to wrap the above into an overall ```if missing or unitialized``` and then do inner ```if```s…

test-requirements.txt
```diff
@@ -1,3 +1,4 @@
+click
```
bowlofeggs: Don't regular requirements get installed in addition to test requirements? If so, I'd think you don't need to put click here since it's part of the usual requirements…

bowlofeggs (approved):

test-requirements.txt
jeremycline: This is what I was doing before I learned how to tell tox to install the package with various extras, I'll remove it :)

fedmsg/commands/check.py
jeremycline: Good call, thanks!

fedmsg/commands/check.py
jeremycline: It actually does report all of them in the for loop above, this is mostly about explaining the reason it exited non-zero.

PR #417, opened 2017-06-07
Set Travis CI to run via tox

jeremycline:
This unifies the tox test environment with the CI environment so that users can run the same test environment CI does.

Additionally, it switches the test runner to pytest (since nose is dead) and generates a coverage report.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -15,7 +15,8 @@ whitelist_externals =
     rm
 commands =
     rm -rf htmlcov coverage.xml
-    python setup.py nosetests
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -1,18 +1,43 @@
...
+commands=
+    mkdir -p _static
+    rm -rf _build
+    sphinx-build -W -b html -d {envtmpdir}/doctrees .  _build/html
+
+[testenv:lint]
+deps =
+    flake8 > 3.0
+commands =
+    python -m flake8 {posargs}
```
bowlofeggs, line 38: Nice

bowlofeggs (approved):

PR #419, opened 2017-06-15
Use irc logger instead of moksha.hub.

ralphbean:
This will make debugging much easier.  Currently, if connections fail you see
"[moksha.hub] Could not connect".  After this change you should see
"[fedmsg.consumers.ircbot] Could not connect." which gives much better
information.

### DIFF
--- a/fedmsg/consumers/ircbot.py
+++ b/fedmsg/consumers/ircbot.py
@@ -166,26 +166,26 @@ class Fedmsg2IRCFactory(protocol.ClientFactory):
             # If we're joining 12 channels, join one of them first.  Once
             # joining, wait five seconds and start joining the second one.
             # That one should trigger joining the third one...
-            self.log.info("%s scheduling conn for next client" % self.nickname)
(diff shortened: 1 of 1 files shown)


jeremycline:
The change itself looks fine to me (all the logging needs similar treatment), but unfortunately it looks like there's no test coverage for that module so the diff coverage is 0%. Could you write some basic tests for that module? If you don't have time that's fine, but it'll be a while before I can get to it.

ralphbean:
> Could you write some basic tests for that module?

Sorry, I'm slammed.  :(

ralphbean:
> Could you write some basic tests for that module?

Started in #432.

jeremycline:
Could you rebase onto develop to pick up the newest tests?

ralphbean:
> Could you rebase onto develop to pick up the newest tests?

Yes.  Done.  Thanks.  :)

jeremycline (approved):
As much as I'd like test coverage for this module, I would prefer to make a release and this is a fairly safe change.

Issue #420, opened 2017-06-16
Rework the CA settings to not use a cache or download from the Internet


jeremycline:
It doesn't make a whole lot of sense for a CA to require being periodically fetched from the Internet (via HTTPS, granted). If your CA is compromised you have a lot of problems and you're not going to wait for the cache to expire to rotate your CA (I assume). It makes much more sense to just configure a filesystem path to your CA and update that as necessary with some configuration management tool. What's more annoying is there's currently no way to do that since you _must_ provide a valid HTTP URL and if the file is older than the cache value you set, it'll be fetched and saved to your cache location.

PR #421, opened 2017-06-16
Use cryptography and pyOpenSSL instead of M2Crypto

### LINKED ISSUE #411: fedmsg-0.18.3: tests fail in crypto
Trying to build fedmsg in rawhide fails in tests. 

```
Does fedmsg.config._recursive_update produce the expected result? ... ok
test_send_message (fedmsg.tests.test_core.TestCore)
send_message is deprecated ... ok
test_invalid_validator (fedmsg.tests.test_crypto_switching.TestCryptoSwitching)
Try to verify an x509 message when only gpg is allowed. ... ok
test_failed_validation (fedmsg.tests.test_crypto_x509.TestCryptoX509)
Try to fail validation. ... /var/tmp/rpm-tmp.FX8uYf: line 31: 10522 Segmentation fault      (core dumped) PYTHONPATH=$(pwd) python setup.py test
RPM build errors:
```

dmesg on the builder has: 

[Sat May 27 18:32:29 2017] python[10522]: segfault at 49 ip 00000000f6e09585 sp 00000000ffd29d00 error 4 in libcrypto.so.1.1.0e[f6c85000+22f000]

Perhaps we should move to x509_ng which doesn't use mcrypto, but python-crypto

jeremycline:
M2Crypto is Python 2 only and is in maintenance mode. Although pyOpenSSL
is _also_ in maintenance mode, it supports Python 3 and relies on
cryptography for its OpenSSL bindings. Currently the only thing
cryptography is missing (that pyOpenSSL has) is certificate chain
validation. Once that's available in cryptography, we can drop
pyOpenSSL.

Unfortunately, we still need to support M2Crypto because the current
versions of cryptography and pyOpenSSL in EL7 is below 1.6 and 16.1.0
respectively.

This patch uses cryptography and pyOpenSSL if available and falls back
to M2Crypto.

fixes #411


fedmsg/crypto/utils.py
```diff
@@ -0,0 +1,104 @@
...
+def validate_policy(topic, signer, routing_policy, nitpicky=False):
+    """
+    Checks that the sender is allowed to emit messages for the given topic.
+
+    Args:
+        topic (str): The message topic the ``signer`` used when sending the message.
+        signer (str): The Common Name of the certificate used to sign the message.
+
+    Returns:
+        bool: True of the policy defined in the settings allows the signer to send
```
bowlofeggs: True *if* the policy…

fedmsg/crypto/utils.py
```diff
@@ -0,0 +1,104 @@
...
+            messages on ``topic``.
+    """
+    if topic in routing_policy:
+        # If so.. is the signer one of those permitted senders?
+        if signer in routing_policy[topic]:
+            # We are good.  The signer of this message is explicitly
+            # whitelisted to send on this topic in our config policy.
+            return True
+        else:
+            # We have a policy for this topic and $homeboy isn't on the list.
```
bowlofeggs, line 29: Haha

fedmsg/crypto/utils.py
```diff
@@ -0,0 +1,104 @@
...
+            # routing_policy for the topic of this message.. and *nobody*
+            # gets in without a pass.  That means that we fail the message.
+            _log.error("Authorization/routing_policy underspecified.")
+            return False
+        else:
+            # We are *not* in nitpicky mode.  We don't have an entry in the
+            # routing_policy for the topic of this message.. but we don't
+            # really care.
+            _log.info('No routing policy defined for "{t}" and routing_nitpicky '
+                      'is False so the message is being treated as authorized.'.format(t=topic))
```
bowlofeggs: This might make sense at warning level.

fedmsg/crypto/utils.py
```diff
@@ -0,0 +1,104 @@
...
+            return True
+
+
+def load_remote_cert(location, cache, cache_expiry, tries=0, **config):
+    """ Get a fresh copy from fp.o/fedmsg/crl.pem if ours is getting stale.
+
+    Return the local filename.
```
bowlofeggs: I recommend documenting the arguments and the return type.

fedmsg/crypto/utils.py
```diff
@@ -0,0 +1,104 @@
...
+        except OSError:
+            # Neither file exists
+            modtime = 0
+
+    if (
+        (not modtime and not cache_expiry) or
+        (cache_expiry and time.time() - modtime > cache_expiry)
+    ):
+        try:
+            response = requests.get(location)
```
bowlofeggs, line 89: Maybe put a timeout on this call to ```get()```.

fedmsg/crypto/utils.py
```diff
@@ -0,0 +1,104 @@
...
+            # If we couldn't write to the specified cache location, try a
+            # similar place but inside our home directory instead.
+            cache = alternative_cache
+            usr_dir = '/'.join(cache.split('/')[:-1])
+
+            if not os.path.isdir(usr_dir):
+                os.makedirs(usr_dir)
+
+            with open(cache, 'w') as f:
+                f.write(response.content)
```
bowlofeggs, line 111: Note that this could also raise an IOError, but that's probably OK.

fedmsg/crypto/x509.py
```diff
@@ -196,87 +196,18 @@ def fail(reason):
...
-        except OSError:
-            # Neither file exists
-            modtime = 0
-
-    if (
-        (not modtime and not cache_expiry) or
-        (cache_expiry and time.time() - modtime > cache_expiry)
-    ):
-        try:
-            response = requests.get(location)
```
bowlofeggs, line 258: Oh I see - you just moved this code block. I'd still recommend that timeout though… (maybe a separate PR or commit if you prefer - up to you).

fedmsg/crypto/x509_ng.py
```diff
@@ -19,256 +19,215 @@
...
-    rsa_public = cert.get_pubkey().get_rsa()
-    digest = M2Crypto.EVP.MessageDigest('sha1')
-    digest.update(fedmsg.encoding.dumps(message))
     try:
-        if not rsa_public.verify(digest.digest(), signature):
-            raise M2Crypto.RSA.RSAError("RSA signature failed to validate.")
-    except M2Crypto.RSA.RSAError as e:
-        return fail(str(e))
+        _validate_signing_cert(ca_certificate, certificate, crl)
+    except X509StoreContextError as e:
```
bowlofeggs, line 180: I presume this step will catch expired certificates and non-properly signed certificates?

fedmsg/tests/crypto/test_x509.py
```diff
@@ -0,0 +1,216 @@
+# This file is part of fedmsg.
+# Copyright (C) 2012 - 2014 Red Hat, Inc.
```
bowlofeggs: 2017?

fedmsg/tests/crypto/test_x509.py
```diff
@@ -0,0 +1,216 @@
...
+        self.config['routing_nitpicky'] = True
+        signed = self.sign({'topic': 'mytopic', 'message': 'so secure'}, **self.config)
+        self.assertFalse(self.validate(signed, **self.config))
+
+    def test_valid_policy(self):
+        """Assert when the policy is valid, True is returned."""
+        self.config['routing_policy'] = {'mytopic': ['shell-app01.phx2.fedoraproject.org']}
+        self.config['routing_nitpicky'] = True
+        signed = self.sign({'topic': 'mytopic', 'message': 'so secure'}, **self.config)
+        self.assertTrue(self.validate(signed, **self.config))
```
bowlofeggs, line 157: How about a ```test_signed_by_expired_cert``` and ```test_signed_by_expired_ca``` (bad name, but what I mean is the signing cert is signed by an expired CA) as well?

fedmsg/tests/crypto/test_x509.py
```diff
@@ -0,0 +1,216 @@
...
+
+from fedmsg import crypto  # noqa: E402
+
+
+SSLDIR = os.path.abspath(
+    os.path.join(os.path.dirname(__file__), '../test_certs/keys/'))
+
+
+@skipIf(not (_m2crypto or _cryptography), "Neither M2Crypto nor Cryptography available")
+class X509BaseTests(TestCase):
```
bowlofeggs, line 49: I like this pattern of making a superclass to get the same tests run with the different implementations. Nice!

fedmsg/tests/crypto/test_x509.py
```diff
@@ -0,0 +1,216 @@
...
+    """Tests that use cryptography for signing and validate with m2crypto."""
+
+    def setUp(self):
+        self.config = {
+            'ssldir': SSLDIR,
+            'certname': 'shell-app01.phx2.fedoraproject.org',
+            'ca_cert_cache': os.path.join(SSLDIR, 'ca.crt'),
+            'ca_cert_cache_expiry': 1497618475,  # Stop fedmsg overwriting my CA, See Issue 420
+
+            'crl_location': "http://threebean.org/fedmsg-tests/crl.pem",
```
bowlofeggs, line 226: Do these tests work offline? If not, I see that you have a ```fedmsg/tests/test_certs/keys/crl.pem``` that you might could use instead…

bowlofeggs (approved):
This looks reasonable to me. Note however that I lack familiarity with the cryptography and pyopenssl libraries, so I am not able to comment on whether they were used correctly or not. Thus, I recommend getting a second opinion from someone who is familiar with those APIs before merging. On the other hand, I did like the tests you wrote and they helped give me confidence in the code ☺

fedmsg/crypto/x509.py
jeremycline: I was actually thinking of doing a separate PR that changes this API a lot (one reason I didn't bother documenting it when I moved it). It does quite a few surprising things which I will probably address in the fix for #420.

fedmsg/crypto/x509_ng.py
jeremycline: Yup, it'll raise for that as well as revoked certificates and expired CRLs

fedmsg/tests/crypto/test_x509.py
jeremycline: They do, but only because I set the cache expiry to be ~47 years. There's no way to have the cert not downloaded from the Internet eventually and it defaults to the Fedora CA and CRL so even if you don't provide a URL it'll happen (all of which I want to change, I just haven't decided how yet).

fedmsg/tests/crypto/test_x509.py
jeremycline: Good idea! I have made it so, Number One.

fedmsg/crypto/utils.py
```diff
@@ -0,0 +1,113 @@
...
+                         'False so the message is being treated as authorized.'.format(t=topic))
+            return True
+
+
+def load_remote_cert(location, cache, cache_expiry, tries=0, **config):
+    """Get a fresh copy from fp.o/fedmsg/crl.pem if ours is getting stale.
+
+    Return the local filename.
+
+    .. note:: This is not a public API and is subject to change.
```
puiterwijk, line 60: Maybe start the name with _ then?

fedmsg/crypto/x509.py
```diff
@@ -19,46 +19,50 @@
...
 
 if six.PY3:
     long = int
 
 
-def sign(message, ssldir=None, certname=None, **config):
+def _disabled_sign(*args, **kwargs):
+    """A fallback function that emits a warning when no crypto is being used."""
+    warnings.warn('Message signing is disabled because "cryptography" and '
+                  '"pyopenssl" or "m2crypto" are not available.')
```
puiterwijk, line 49: Perhaps add parentheses to indicate the and/or relations?

fedmsg/crypto/x509_ng.py
```diff
@@ -19,256 +19,221 @@
...
             password=None,
             backend=default_backend()
         )
 
-    signer = rsa_private.signer(
+    signature = rsa_private.sign(
+        fedmsg.encoding.dumps(message).encode('utf-8'),
         asymmetric.padding.PKCS1v15(),
-        hashes.SHA1()
+        hashes.SHA1(),
```
puiterwijk, line 85: While it's fine right now, it would be great if we could start migrating to new hashes, like sha256/sha512.
Maybe add an indicator which is used before the next version, so we can switch easier in the future?

fedmsg/crypto/x509_ng.py
```diff
@@ -19,256 +19,221 @@
...
+        )
+    except InvalidSignature as e:
+        _log.error('message [{m}] has an invalid signature: {e}'.format(
+            m=message, e=str(e)))
+        return False
 
-    # Load our policy from the config dict.
+    # Step 4, check that the certificate is permitted to emit messages for the
+    # topic.
+    common_name = crypto_certificate.subject.get_attributes_for_oid(x509.oid.NameOID.COMMON_NAME)
```
puiterwijk, line 200: You should not use the common name ideally, this is being deprecated by the CA/Browser Forum baseline requirements.
While we are not reliant on those, I think ideally we should follow their move to the Subject Alternative Name.
This does mean there may be multiple records, and we might need to check all SAN values in the authorization step.
For the time being, for compatibility, we might add the Common Name to the list of SANs for checking purposes though and issue a warning if the CN was not in any SAN values.

fedmsg/crypto/x509_ng.py
```diff
@@ -19,256 +19,221 @@
...
+    common_name = common_name[0]
     routing_policy = config.get('routing_policy', {})
+    nitpicky = config.get('routing_nitpicky', False)
+    return utils.validate_policy(
```
puiterwijk, line 204: An idea would be a return any(...) over all the SANs.

fedmsg/tests/crypto/test_x509.py
```diff
@@ -0,0 +1,237 @@
...
+class X509BaseTests(TestCase):
+
+    def setUp(self):
+        self.config = {
+            'ssldir': SSLDIR,
+            'certname': 'shell-app01.phx2.fedoraproject.org',
+            'ca_cert_cache': os.path.join(SSLDIR, 'ca.crt'),
+            'ca_cert_cache_expiry': 1497618475,  # Stop fedmsg overwriting my CA, See Issue 420
+
+            'crl_location': "http://threebean.org/fedmsg-tests/crl.pem",
```
puiterwijk, line 58: Ideally we would not depend on this website... Maybe use a file:// ?

puiterwijk (approved):
Overall this looks good, with some future suggestions.

fedmsg/crypto/x509_ng.py
jeremycline: I considered this, but I didn't want to make that change alongside all the other stuff that's already happening. I thought I filed an issue, but it seems like I hadn't, so I filed #438

puiterwijk: That makes perfect sense. This was just an idea now that you got me rereading the code :).

fedmsg/crypto/x509_ng.py
jeremycline: Yeah, I wanted to do this, but the M2Crypto version is only using the CN. I will definitely look into moving both of them to SAN (#439).

fedmsg/tests/crypto/test_x509.py
jeremycline: Most unfortunately, the way the code currently works it doesn't support file://. I set the cache expiry to be sometime in the mid 2030s so it should never actually try to fetch from that URL. I plan to re-write all the cert fetching code in the near future.

fedmsg/crypto/x509_ng.py
puiterwijk: The concrete sentence about Common Name from the CA/Browser Baselines requirements v1.4.5 (which is currently the latest stable, source https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.4.5.pdf):

````
Certificate Field: subject:commonName (OID 2.5.4.3)   
Required/Optional: Deprecated (Discouraged, but not prohibited) 
````

PR #430, opened 2017-06-21
Make the ircbot url shortener configurable.

ralphbean:
Someone can provide something like this in their configuration to make link
shortening work in... some other way::

```python
config = dict(
    irc=[dict(
            make_short=lambda url: "http://awesome",
    )],
)
```

### DIFF
--- a/fedmsg/consumers/ircbot.py
+++ b/fedmsg/consumers/ircbot.py
@@ -66,6 +66,12 @@ mirc_colors = {
 }
 
 
+def _default_link_shortener(url):
+    dagd = 'http://da.gd/s'
+    resp = requests.get(dagd, params=dict(url=url))
+    return resp.text.strip()
+
+
 def ircprettify(title, subtitle, link="", config=None):
     def markup(s, color):
         return "\x03%i%s\x03" % (mirc_colors[color], s)
@@ -307,9 +313,10 @@ class IRCBotConsumer(FedmsgConsumer):
                     link = fedmsg.meta.msg2link(msg, **self.hub.config)
 
                 if link and short:
-                    dagd = 'http://da.gd/s'
-                    resp = requests.get(dagd, params=dict(url=link))
-                    link = resp.text.strip()
+                    if callable(short):
+                        link = short(link)
+                    else:
+                        link = _default_link_shortener(link)
 
                 return ircprettify(
                     title=title,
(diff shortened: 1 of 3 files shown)


fedmsg/consumers/ircbot.py
```diff
@@ -307,9 +313,10 @@ def prettify(self, topic, msg, pretty=False, terse=False, short=False):
                     link = fedmsg.meta.msg2link(msg, **self.hub.config)
 
                 if link and short:
-                    dagd = 'http://da.gd/s'
-                    resp = requests.get(dagd, params=dict(url=link))
-                    link = resp.text.strip()
+                    if callable(short):
```
jeremycline, line 320: Is the intention here to use some other shortener service? I ask because this approach is probably going to make error handling tricky since any sort of exception could be raised if someone opts to not use requests.

jeremycline:
I haven't really investigated this particular consumer before, but I'm surprised it works. It seems to be using Twisted features outside the reactor thread.

fedmsg/consumers/ircbot.py
```diff
@@ -66,6 +66,12 @@
 }
 
 
+def _default_link_shortener(url):
+    dagd = 'http://da.gd/s'
+    resp = requests.get(dagd, params=dict(url=url))
```
jeremycline: This should definitely have a ``timeout`` kwarg applied to it.

ralphbean: Can do.

fedmsg/consumers/ircbot.py
ralphbean: Yes, that's the intention.

As a user - I'd like to still use `requests`, but just a different shortener service.

jeremycline: Perhaps it would be better to just let a user configure the service URL and require that the service return the shortened link as the response body to a GET with the ``url`` parameter. Will that meet your needs?

jeremycline (changes requested):
Could you please add tests (100% diff test coverage is required) and documentation on the new configuration option?

ralphbean:
> Perhaps it would be better to just let a user configure the service URL and require that the service return the shortened link as the response body to a GET with the url parameter. Will that meet your needs?

Unfortunately, no.  :(  I wish!

The shortener that I want (need) to use does accept GET requests, but it doesn't take a url parameter.

It is of the form:

https://shortener.com/new?the_url_goes_here

If it was like

https://shortener.com/new?url=the_url_goes_here

Then your approach would work.

ralphbean:
> Could you please add tests (100% diff test coverage is required) and documentation on the new configuration option?

Tests and docs attached.

jeremycline (approved):
Thanks!

Issue #425, opened 2017-06-16
Simplify the fedmsg configuration


jeremycline:
This is up for discussion, but I'd like to propose that we start migrating to a simpler configuration format. I think we can do this in a backwards-compatible way (where we support both the new and old method and eventually remove the old method).

What I'd like is:

1. Use the TOML format loaded with pytoml

2. Load from ~/.config/fedmsg/config.toml. If it's not present, load from /etc/fedmsg/config.toml. Don't do any fancy merging. Allow for configuration to be specified via the CLI as well, which takes precedence over both those locations.
 
3. Configuration is lazy-loaded and not actually read until the configuration dictionary is actually accessed.

4. Ensure all the defaults are user-friendly and not Fedora-specific.

To use the configuration you just do

```
>>> from fedmsg import configuration
>>> configuration['zmq_enabled']
True
```

We can support the old configuration loading (with deprecation warnings) for a while yet and then drop it when the world is ready for the future.

pypingou:
Do note that this is going to impact quite some apps, in our control and out of it

jeremycline:
It'll be backwards compatible and people will have plenty of time to migrate

glensc:
i like the fact that configuration is in separate files. that allows distribution (rpm) package provide default config. and less pain for sysadmins to merge huge config when some new component is added (distro package with it's own config).

jeremycline:
That's a good point.

Perhaps a better scheme would be:

If a file is specified on the CLI, use that (and only that) configuration. Otherwise, load all configuration from /etc/fedmsg/conf.d/. If there are duplicate keys in any of the files, report it as an error to the user.

PR #431, opened 2017-06-22
add topic_prefix example

glensc:
currently it's not visible, [hardcoded value](https://github.com/fedora-infra/fedmsg/blob/0.18.3/fedmsg/config.py#L55) from codebase

### DIFF
--- a/fedmsg.d/base.py
+++ b/fedmsg.d/base.py
@@ -20,6 +20,9 @@
 import os
 
 config = dict(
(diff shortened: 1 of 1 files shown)


jeremycline (approved):
Thanks @glensc!

To be honest, I'd prefer it if it wasn't Fedora specific, but I bet plenty of things rely on that. I'll make a note on #425 to fix all the defaults :shipit:

PR #432, opened 2017-06-22
Pass along headers if present.

ralphbean:
These may be useful to filters or fedmsg.meta routines.

### DIFF
--- a/fedmsg/consumers/ircbot.py
+++ b/fedmsg/consumers/ircbot.py
@@ -342,6 +342,11 @@ class IRCBotConsumer(FedmsgConsumer):
         log.debug("Got message %r" % msg)
         topic, body = msg.get('topic'), msg.get('body')
 
+        # Pass along headers if present.  May be useful to filters or
+        # fedmsg.meta routines.
+        if 'headers' in msg:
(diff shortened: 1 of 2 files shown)


fedmsg/consumers/ircbot.py
```diff
@@ -342,6 +342,11 @@ def consume(self, msg):
         log.debug("Got message %r" % msg)
         topic, body = msg.get('topic'), msg.get('body')
 
+        # Pass along headers if present.  May be useful to filters or
+        # fedmsg.meta routines.
+        if 'headers' in msg
```
jeremycline: You missed a colon here.

jeremycline (changes requested):
There's a syntax error. I'd also really like to see test coverage for this change.

fedmsg/consumers/ircbot.py
ralphbean: Doh, thanks.  Rebased.

ralphbean:
> I'd also really like to see test coverage for this change.

Added in 951d1aa.

jeremycline (approved):
Looks good, thanks!

:shipit:

PR #434, opened 2017-06-27
Mark the --daemon option as deprecated

### LINKED ISSUE #423: Deprecate the daemonize option for commands
See https://github.com/fedora-infra/fedmsg/pull/398 for details, but we should just let users daemonize commands via systemd (or their init system of choice).

This issue is for adding deprecation warnings when that option is used. I propose we remove them in version 0.20 of fedmsg.

jeremycline:
The setup.py also depended on the wrong PyPI package.

fixes #423

### DIFF
--- a/fedmsg/commands/__init__.py
+++ b/fedmsg/commands/__init__.py
@@ -81,6 +81,9 @@ class BaseCommand(object):
         except:
             from daemon.pidlockfile import PIDLockFile
 
+        msg = ('The use of the "--daemon" flag is deprecated and will be removed in '
+               'fedmsg-0.20.0. Use your init system to run fedmsg commands as daemons.')
+        warnings.warn(msg, category=UserWarning)
         pidlock = PIDLockFile('/var/run/fedmsg/%s.pid' % self.name)
 
         pid = pidlock.read_pid()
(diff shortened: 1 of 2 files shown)


fedmsg/commands/__init__.py
```diff
@@ -81,6 +81,9 @@ def _daemonize(self):
         except:
             from daemon.pidlockfile import PIDLockFile
 
+        msg = ('The use of the "--daemon" flag is deprecated and will be removed in '
+               'fedmsg-0.20.0. Use your init system to run fedmsg commands as daemons.')
+        warnings.warn(msg, category=UserWarning)
```
bowlofeggs, line 86: I guess this code isn't covered by tests, which is why this is causing coverage to go down. I would recommend writing a test to hit this and maybe make some assertions about the method.

bowlofeggs (approved):

fedmsg/commands/__init__.py
jeremycline: Yeah, I was actually too lazy to run the tests locally to see if the coverage would go down so I just made a PR :(

jeremycline:
It shames me, but I think I'm going to merge this since right after 0.19.0 is released I'm going to delete all this code anyway.

bowlofeggs:
On Tue, 2017-06-27 at 10:58 -0700, Jeremy Cline wrote:
> It shames me, but I think I'm going to merge this since right after
> 0.19.0 is released I'm going to delete all this code anyway.

Works for me.

PR #437, opened 2017-06-28
Pass along headers for *all* consumers, not just the irc bot.

ralphbean:
This is on top of #432.  We realized we had the same problem with datanommer that we saw in the irc bot.

### DIFF
--- a/fedmsg/consumers/__init__.py
+++ b/fedmsg/consumers/__init__.py
@@ -246,6 +246,12 @@ class FedmsgConsumer(moksha.hub.api.consumer.Consumer):
         except RuntimeWarning as e:
             self.log.warn("Received invalid message {0}".format(e))
             return
+
+        # Pass along headers if present.  May be useful to filters or
+        # fedmsg.meta routines.
+        if 'headers' in message and 'body' in message:
+            message['body']['headers'] = message['headers']
+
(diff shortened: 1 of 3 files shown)


mikebonnet:
+1 Works for me.

jeremycline (approved):
This is fine with me, although I'm not crazy about stuffing things into the message body.

One thing I've been mulling on since I worked on the crypto module rewrite is that is would probably be nice to have a message envelope. Placing the signature inside the data structure that was just signed makes it difficult to verify and isn't really relevant to the user. They just want to know it's a valid message. Maybe this "envelope" would be a good place for headers in the long term.

ralphbean:
> Placing the signature inside the data structure that was just signed makes it difficult to verify and isn't really relevant to the user.

Agreed.  That was a mistake in hindsight.

ralphbean:
@jeremycline, I don't understand why codecov is blocking the merge here.  Any insights?

jeremycline:
I set it so coverage was not allowed to drop from a pull request. It looks like the coverage report isn't including ``fedmsg/consumer/__init__.py`` for some reason, so it just sees the hit count going down. Maybe some weird rounding is also occurring. I've turned off the requirement for project coverage increasing for the moment.

PR #440, opened 2017-07-03
Add classifiers to the setup.py

jeremycline:


setup.py
```diff
@@ -122,6 +122,17 @@
     author_email='rbean@redhat.com',
     url='https://github.com/fedora-infra/fedmsg/',
     license='LGPLv2+',
+    classifiers=[  # https://pypi.python.org/pypi?%3Aaction=list_classifiers
+        'Development Status :: 3 - Alpha',
```
pypingou: fedmsg in alpha state?

jeremycline: It's still releasing 0.y.z, which, according to semvar means

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

We may be using it in production, but I think there are plenty of APIs in fedmsg that should change before we claim it's stable.

jeremycline: After some discussion on IRC, there appears to be some disagreement on this particular classifier. I'd like expand upon my justification. Those who don't agree with my justification, please make a counter-proposal. Let's have a discussion about this.

All these justifications assume that fedmsg follows semantic versioning 2.0.0 (http://semver.org/). @ralphbean, since you're the original author, is this assumption false?

As stated above, it's currently releasing under the a 0.y.z scheme which means the API is not stable and could change at any time.

Secondly, there are several API changes I would like to enact prior to declaring fedmsg's API as stable. These include (but are not necessarily limited to): 

1. Proper handling of text vs binary data. This is particularly important since ZeroMQ deals with binary data, not text. Some APIs should start requiring ``unicode``/``str`` objects and rejecting binary data (e.g. the ``fedmsg.publish`` API) and some APIs should start returning unicode objects.

2. I'd prefer to drop moksha and move the functionality we need into fedmsg proper.

3. I'd like to move away from a Python file format for configuration.

4. I'd like to change the on-the-wire format of fedmsgs. I believe the documentation includes using pyzmq directly. We need to decide whether or not this will be a public API.

5. I'd like to unify the CLI with click.

All that to say, I'd like to take advantage of the alpha status of the API.

pypingou: I believe most of our apps have been following https://fedoraproject.org/wiki/Infrastructure/AppBestPractices#Version_numbering rather than entire semantic versioning.

So all you want to do would be doable, but would require a bump of the major version, from 0 to 1.

ralphbean: > All these justifications assume that fedmsg follows semantic versioning 2.0.0 (http://semver.org/). @ralphbean, since you're the original author, is this assumption false?

In theory, no - the assumption is correct.  In practice, yes - the assumption is incorrect.

Like @pypingou pointed out, there are more than a few apps depending on the existing API regardless of whether or not a `1.x` promise was made.

I'd advise treating `0.x` as a stable line and any API breaking changes could go into a new `1.x` line.  You could even jump straight from `0.x` to `2.x` to make the jump obvious to users.

bowlofeggs: The version numbering wiki thing sounds pretty similar to semver to me. It's not exact and doesn't go into as much detail, but the basic idea sounds the same to me. FWIW, I've been trying to just follow semver for Bodhi because I didn't know about that document, though I also don't think I've done anything that that would "disagree" with it either.

Another idea to confuse things even more: what if we just go ahead and mark fedmsg as being a 1.0 and call it stable? Then, we can always make a 2.0 when we are ready to change the API. I don't think that would be practically different than keeping it as a 0.y for a long time and then changing it for a 1.0 or a 2.0. Tons of apps are currently relying on its API, so going ahead and calling it a 1.0 could be a nice statement for all the other apps. And we can still make 2.0 when we need to change it. Just an idea.

jeremycline: If people are more comfortable with the X release being bumped than they are with the 0.y release being bumped, that's fine with me.

However, I'd still like the next release to be 0.19 so I have a chance to get the documentation in order and determine what API I'm declaring stable (e.g. is the on-the-wire format part of the API? What Python modules, exactly?).

bowlofeggs: I think it'd be useful to get fedmsg into the semver phase where the X number means the API is changing - that's a nice handy way to indicate to the users that they need to watch out ☺

0.19 sounds fine to me.

sayanchowdhury: Yeah, that's a fine plan to follow.

jeremycline: Okay, that's what we'll do, then. In the mean time, unless anyone has any objections, I'd like the Python classifier to match the version. I'll change it to "Stable" when I release 1.0.0. @pypingou are you okay with this plan?

pypingou: I'd suggest not including that classifier at the moment and add it in 1.0.0 as you proposed.

jeremycline: Very well.

pypingou (approved):
LGTM

PR #446, opened 2017-07-13
The m2crypto tests are skipped when cryptography is missing

jeremycline:
The intention was to run the m2crypto tests only when m2crypto is
present, but the actual check was for cryptography. Oops.


cverna (approved):
LGTM

PR #448, opened 2017-07-13
Remove the dependency on nose

jeremycline:
Now that we use pytest, we should not depend on nose.


pypingou:
Looks fine to me

PR #449, opened 2017-07-13
Add Python 3 support to fedmsg

jeremycline:
This adds Python 3 support to fedmsg. Tests are run against 3.4+.


pypingou:
Nice and clean, looks fine to me! :)

jeremycline:
I believe this also depends on #448 since with nose the tests hang on Python 3. I was working off that branch and rebased before I made this PR. I don't know _why_ nose causes the tests to hang on Python 3, though :(

jeremycline:
Okay, I had to tweak a few more things to get CI passing, so it probably wouldn't hurt for this to get a second look.

pypingou:
Still looking good to me 👍

PR #452, opened 2017-07-18
Add missing test requirement - click

frenzymadness:
CliRunner from click.testing is widely used in test_commands.py but it's missing in test-requirements.txt.


jeremycline (approved):
Ah, good catch!

PR #456, opened 2017-08-03
Allow validating unicode signatures.

ralphbean:
I just noticed this in the logs of program driven by `fedmsg.tail_messages`.
The signature was unicode, and so this type check failed every time.

This seems to have been inadvertently broken in f0553d285e586dc48f1a244b81f1517551d469ee ?

### DIFF
--- a/fedmsg/crypto/x509.py
+++ b/fedmsg/crypto/x509.py
@@ -113,7 +113,7 @@ def _m2crypto_validate(message, ssldir=None, **config):
     for field in ['signature', 'certificate']:
         if field not in message:
             return fail("No %r field found." % field)
-        if not isinstance(message[field], str):
+        if not isinstance(message[field], basestring):
             return fail("msg[%r] is not a string" % field)
 
     # Peal off the auth datums
--- a/fedmsg/crypto/x509.py
+++ b/fedmsg/crypto/x509.py
@@ -37,7 +37,7 @@ import fedmsg.crypto  # noqa: E402
 import fedmsg.encoding  # noqa: E402
 
 
-log = logging.getLogger(__name__)
+_log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


fedmsg/crypto/x509.py
```diff
@@ -113,7 +113,7 @@ def fail(reason):
     for field in ['signature', 'certificate']:
         if field not in message:
             return fail("No %r field found." % field)
-        if not isinstance(message[field], str):
+        if not isinstance(message[field], basestring):
```
jeremycline: ``basestring`` is undefined in Python 3. While this code will never actually run in Python 3 since it's the M2Crypto implementation, this makes the API unclear and it should either be ``six.text_type`` (unicode and str in Python 2 and Python 3, respectively) or ``six.binary_type`` (str and bytes in Python 2 and Python 3, respectively).

Since we don't know the encoding at this point, I believe it should be ``six.text_type``.

jeremycline (changes requested):
Thanks for the PR! 

This change needs tests to demonstrate the issue before it can be merged.

fedmsg/crypto/x509.py
ralphbean: That's legit.

The underlying issue is (I think) that some routines pass unicode signatures in here, while others pass byte-strings.

I guess I'm with you.. that we should enforce "unicode" as the api with ``six.text_type`` and then go looking for any places that may erroneously call this with byte strings.

puiterwijk:
Quick glance says that this looks okay to me.

ralphbean:
> Quick glance says that this looks okay to me.

Same.

jeremycline (approved):
:shipit:

jeremycline:
This patch was backported to 0.19.1

PR #458, opened 2017-08-07
Improved tox configuration

frenzymadness:
Hi.
I think that these two small handy changes can improve the comfort of a developer.
* Time needed to execute tests can be reduced when you need to execute only subset of them. 
* You need to pass environment variable FEDMSG_NETWORK if you want to run network test.


jeremycline (approved):
Great, thanks!

:shipit:

PR #459, opened 2017-08-07
Lower the number of skipped test

frenzymadness:
Hi.
I've added some expected variables to classes inherited from Base in test_meta to lower the number of skipped tests. This change lowers the number of skipped test from 98 to 65. I also sorted it alphabetically and move one-line variables to the top.
I think that is good to keep the number of skipped test as small as possible and this tests at least check the proper return value/type.


jeremycline (approved):
Looks good, thanks!

PR #461, opened 2017-08-09
In py2, the signature is a unicode object

puiterwijk:
Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/fedmsg/crypto/x509.py
+++ b/fedmsg/crypto/x509.py
@@ -113,7 +113,8 @@ def _m2crypto_validate(message, ssldir=None, **config):
     for field in ['signature', 'certificate']:
         if field not in message:
             return fail("No %r field found." % field)
-        if not isinstance(message[field], str):
+        if not isinstance(message[field], str) and \
+                (six.PY3 or not isinstance(message[field], unicode)):
             return fail("msg[%r] is not a string" % field)
 
     # Peal off the auth datums


puiterwijk:
This really needs to be fixed before the mass updates tomorrow, as this break fedmsg-hub thoroughly.

puiterwijk:
Okay, this was actually already fixed. We still REALLY need a new release with the basestring fix, as the version in EPEL has isinstance(..., str).

fedmsg/crypto/x509.py
```diff
@@ -113,7 +113,8 @@ def fail(reason):
     for field in ['signature', 'certificate']:
         if field not in message:
             return fail("No %r field found." % field)
-        if not isinstance(message[field], str):
+        if not isinstance(message[field], basestring) and \
```
pypingou: basestring won't work with PY3

puiterwijk: basestring is the current code....

pypingou: Isn't the current code using ``str``?

puiterwijk: So it is, I did a terribly job rebasing. Then we *do* need this patch, just reworked.

jeremycline:
This is a duplicate of #456, which I'll clean up and test now.

PR #462, opened 2017-08-09
ZMQMessage isn't iterable, use __dict__

relrod:
This was flooding logs with traces after we did updates on 2017-08-09. This is the hotfix I came up with, but if it's fixed another way that's fine with me too. Just wanted to get this out there.

### DIFF
--- a/fedmsg/consumers/__init__.py
+++ b/fedmsg/consumers/__init__.py
@@ -286,7 +286,7 @@ class FedmsgConsumer(moksha.hub.api.consumer.Consumer):
 
         # Pass along headers if present.  May be useful to filters or
         # fedmsg.meta routines.
-        if 'headers' in message.__dict__ and 'body' in message.__dict__:
+        if isinstance(message, dict) and 'headers' in message and 'body' in message:
             message['body']['headers'] = message['headers']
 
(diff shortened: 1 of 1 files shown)


relrod:
It looks like sometimes this method gets called with a dict and other times with a ZMQMessage? Or the tests are wrong. It was definitely a ZMQMessage in the traces we were seeing, but the tests seem to pass it a dict (which is why they failed).

jeremycline:
Oof. I think the tests are wrong, but because the API isn't defined at all I have no idea. I'll do some research on this.

jeremycline:
Okay, none of the moksha APIs are defined or documented and it's a tangled mess of functions accepting "messages" so it's not trivial to say what might break if we suddenly switch from the ZMQMessage class to a plain dict.

My inclination is to either revert the commit that added this and spend some time auditing everything and writing careful tests, or add a check to see if it's a dict before placing headers in the body. @ralphbean, since this is a feature you needed, do you have an opinion one way or the other?

ralphbean:
> add a check to see if it's a dict before placing headers in the body.

This seems like the most practical approach for now, no?

ralphbean (approved):

Issue #463, opened 2017-08-09
after upgrade to 0.19.0: exceptions.TypeError: string indices must be integers


kparal:
When I update from
`fedmsg-0.18.2-1.fc25.noarch`
to
`fedmsg-0.19.0-1.fc25.noarch`
the fedmsg-hub starts failing with this traceback:
```
Traceback (most recent call last):
  File "/usr/lib64/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/usr/lib64/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
  File "/usr/lib64/python2.7/site-packages/twisted/internet/posixbase.py", line 597, in _doReadOrWrite
    why = selectable.doRead()
  File "/usr/lib/python2.7/site-packages/txzmq/connection.py", line 284, in doRead
    log.callWithLogger(self, self.messageReceived, message)
--- <exception caught here> ---
  File "/usr/lib64/python2.7/site-packages/twisted/python/log.py", line 101, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/usr/lib64/python2.7/site-packages/twisted/python/log.py", line 84, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/usr/lib64/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/usr/lib64/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
  File "/usr/lib/python2.7/site-packages/txzmq/pubsub.py", line 72, in messageReceived
    self.gotMessage(message[1], message[0])
  File "/usr/lib/python2.7/site-packages/moksha/hub/zeromq/zeromq.py", line 197, in chain_over_moksha_callbacks
    f(_body, _topic)
  File "/usr/lib/python2.7/site-packages/moksha/hub/zeromq/zeromq.py", line 224, in intercept
    return callback(ZMQMessage(_topic, _body))
  File "/usr/lib/python2.7/site-packages/fedmsg/consumers/__init__.py", line 245, in _consume
    self.validate(message)
  File "/usr/lib/python2.7/site-packages/fedmsg/consumers/__init__.py", line 236, in validate
[...]

PR #464, opened 2017-08-09
Deserialize ZMQMessage bodies when they're unicode

jeremycline:
See #463

### DIFF
--- a/fedmsg/consumers/__init__.py
+++ b/fedmsg/consumers/__init__.py
@@ -26,8 +26,10 @@ import psutil
 import requests
 import threading
 import time
+import warnings
 
 import moksha.hub.api.consumer
+import six
 
 import fedmsg.crypto
 import fedmsg.encoding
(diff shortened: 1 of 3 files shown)


fedmsg/consumers/__init__.py
```diff
@@ -226,11 +227,30 @@ def _make_query(page=1):
...
+
+        This needs to raise an exception, caught by moksha.
+
+        Args:
+            message (dict): The message as a dictionary. This must, at a minimum,
+                contain the 'topic' key with a unicode string value and 'body' key
+                with a dictionary value.
+
+        Raises:
+            RuntimeWarning: If the message is not valid.
```
bowlofeggs: Would ```ValueError``` make sense instead of RuntimeWarning? If not, what about ```RuntimeError``` instead of ```RuntimeWarning```?

Also, you could also document the potential encoding exception if the message fails to decode with the UTF-8 codec.

fedmsg/consumers/__init__.py
```diff
@@ -226,11 +227,30 @@ def _make_query(page=1):
...
             message = message.__json__()
-            if isinstance(message['body'], str):
+            if isinstance(message['body'], six.text_type):
                 message['body'] = json.loads(message['body'])
+            elif isinstance(message['body'], six.binary_type):
+                # Try to decode the message body as UTF-8 since it's very likely
+                # that that was the encoding used. This API should eventually only
+                # accept unicode strings inside messages. If a UnicodeDecodeError
+                # happens, let that bubble up.
+                self.log.error('Message body is not unicode; this is deprecated')
```
bowlofeggs: I recommend using the ```warnings``` module here to issue a deprecation warning, rather than logging an error. After all, using deprecated code isn't actually an error… yet ☺

fedmsg/tests/consumers/test_consumers.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def test_invalid_signature(self):
+        """Assert a RuntimeWarning is raised for topic mismatches."""
+        message = {'topic': 't1', 'body': crypto.sign({'topic': 't1'}, **self.config)}
+        message['body']['signature'] = 'thisisnotmysignature'
+
+        self.assertRaises(RuntimeWarning, self.consumer.validate, message)
+
+    def test_no_topic_in_body(self):
+        """Assert the API accepts dictionary messages."""
```
bowlofeggs: The name of this test and this comment don't seem to match. The comment seems copied from ```test_valid_signature``` above?

fedmsg/tests/test_commands.py
```diff
@@ -270,6 +272,7 @@ class CheckTests(unittest.TestCase):
     """Tests for the :class:`fedmsg.commands.check.CheckCommand`."""
 
     def setUp(self):
+        self.maxDiff = None
```
bowlofeggs: Do you want this change to be here? Might have just been for debugging…

fedmsg/tests/test_commands.py
```diff
@@ -1,4 +1,6 @@
 import resource
```
bowlofeggs, line 22: Not related to this PR, but it might be good to give this file a Copyright block.

bowlofeggs (approved):

fedmsg/consumers/__init__.py
jeremycline: I can think of a number of better exception types, but I'm just documenting the existing interface.

I added the decode error exception though, good catch.

fedmsg/consumers/__init__.py
jeremycline: I guess I'll switch it, but I don't know that deprecated is the right word for this either. It seemed easier than saying "this is probably not going to work in the way you expect and produce garbage data because I don't know what the encoding is".

😞

fedmsg/tests/consumers/test_consumers.py
jeremycline: You caught me, I copy and paste test stubs a lot :(

jeremycline:
This patch was backported to 0.19.1

PR #468, opened 2017-08-16
This flows better...

ralphbean:
.. don't you think?


pypingou:
I agree!

jeremycline (approved):
_Technically_ there should be a comma before the but, but I'll allow it

:shipit:

Issue #418, opened 2017-06-09
Proposal: fedmsg-0.19 is the last release that supports Python 2.6


jeremycline:
I would like to propose that we drop support for Python 2.6 after the next release of fedmsg.

pypingou:
hm, we still have some EL6 boxes that are running fedmsg iirc (FAS comes to mind as an example)

pypingou:
And there may be other customer using fedmsg on old boxes

jeremycline:
I'm not opposed to backporting critical bugfixes to 0.19, but plenty of our dependencies have already dropped 2.6 (Twisted, for example).

jeremycline:
Fixed in #469

PR #469, opened 2017-08-16
Remove Python 2.6 support

jeremycline:
See https://github.com/fedora-infra/fedmsg/issues/418 and https://github.com/mokshaproject/moksha/pull/48.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -13,6 +13,8 @@ Backwards incompatible changes
   `systemd unit files <https://github.com/fedora-infra/fedmsg/tree/1.0.0/initsys>`_
   are available in the git repository (`#470 <https://github.com/fedora-infra/fedmsg/pull/470>`_).
 
+* Python 2.6 is no longer supported (`#469 <https://github.com/fedora-infra/fedmsg/pull/469>`_).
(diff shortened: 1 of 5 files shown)


doc/changelog.rst
```diff
@@ -2,6 +2,15 @@
 Changelog
 =========
 
+1.0.0
+=====
+
+Backwards incompatible changes
+------------------------------
+
+* Python 2.6 is no longer supported
```
pypingou: Will conflict with #470

pypingou (approved):
There will be a conflict with #470 but easy enough to fix.

👍 for me

doc/changelog.rst
jeremycline: I figured this way I'd be _forced_ to write the release note instead of putting it off until I want to tag the release :smile:

jeremycline:
Not sure what the deal is with codecov here, the report I generate locally shows that line as covered and there's no reason I can see that that code only would have been exercised in Python 2.6 alone.

frenzymadness:
Perfect work! Side note: The specfile should be also modified. Should I create a bug in bugzilla?

PR #477, opened 2017-09-05
Make fedmsg replays from datagrepper work in most cases

### LINKED ISSUE #403: fedmsg signature validation appears to fail when messages are played back
I don't actually know what pieces are responsible for playing back fedmsgs, but the-new-hotness failed to validate all the messages it got:
```
[    fedmsg    INFO] Retrieving datagrepper page 2 of 813
[fedmsg.crypto.x509 WARNING] Failed validation.  bad signature
[moksha.hub WARNING] Received invalid message RuntimeWarning('Failed to authn message.',)
...
```

I believe [this](https://apps.fedoraproject.org/datagrepper/id?id=2017-247c500d-657a-4ee8-ae29-9c67a5525973&is_raw=true&size=extra-large) is one of the messages it played back and did not validate. Based on my brief examination of the code, it seems to be failing [here](https://github.com/fedora-infra/fedmsg/blob/develop/fedmsg/crypto/x509.py#L184). This is odd since it's using the certificate it loaded from the message itself, and it seems like that certificate must have been signed by the CA to get that far.

I think this is related to https://github.com/fedora-infra/datanommer/pull/87

jeremycline:
Message playback was not working because the messages did not pass
validation. The reason they didn't pass validation is datanommer mutates
the original fedmsg. This breaks the x509 signature. This change tries
to undo all the changes datanommer makes, but there are several cases
where it is impossible to know what the correct value for a key is so
some messages might still fail signature validation.

The longer term fix is to change how datanommer works so that it does
not alter the message in any way. Additionally, the fedmsg signature
approach could be made much less fragile.

For those that come after me, I am sorry about all the things I did in
this patch.

fixes #403


fedmsg/crypto/utils.py
```diff
@@ -7,6 +7,52 @@
...
+    Datagrepper adds the 'source_name' and 'source_version' keys. If messages happen
+    to use those keys, they will fail message validation. Additionally, a 'headers'
+    dictionary is present on all responses, regardless of whether it was in the
+    original message or not. This is deleted if it's null, which won't be correct in
+    all cases. Finally, datagrepper turns the 'timestamp' field into a float, but it
+    might have been an integer when the message was signed.
+
+    A copy of the dictionary is made and returned if altering the message is necessary.
+
+    I'm so sorry.
```
bowlofeggs, line 24: lol. This does all sound pretty crazy!

fedmsg/crypto/utils.py
```diff
@@ -7,6 +7,52 @@
...
+
+    Args:
+        message (dict): A message to clean up.
+
+    Returns:
+        dict: A copy of the provided message, with the datagrepper-related keys removed
+            if they were present.
+    """
+    if not ('source_name' in message and 'source_version' in message):
```
bowlofeggs, line 33: Perhaps this ought to be an ```or``` instead of ```and```?

fedmsg/crypto/utils.py
```diff
@@ -7,6 +7,52 @@
...
+
+    # Don't mutate the original message
+    message = message.copy()
+
+    del message['source_name']
+    del message['source_version']
+    # datanommer adds the headers field to the message in all cases.
+    # This is a huge problem because if the signature was generated with a 'headers'
+    # key set and we delete it here, messages will fail validation, but if we don't
+    # messages will fail validation if they didn't have a 'headers' key set.
```
bowlofeggs, line 44: So here's another crazy idea: if headers is null and validation fals, should we try adding a null headers back and see if *that* version passes validation?

fedmsg/tests/consumers/test_consumers.py
```diff
@@ -31,12 +32,45 @@
...
+            'dummy': True,
+            'ssldir': SSLDIR,
+            'ca_cert_cache': os.path.join(SSLDIR, 'fedora_ca.crt'),
+            'ca_cert_cache_expiry': 1497618475,  # Stop fedmsg overwriting my CA, See Issue 420
+            'crypto_validate_backends': ['x509'],
+        }
+        self.hub = mock.Mock(config=self.config)
+        self.consumer = DummyConsumer(self.hub)
+
+    def test_backlog_message_validation(self):
```
bowlofeggs, line 59: It'd be good to add a docblock here.

fedmsg/tests/crypto/test_utils.py
```diff
@@ -7,6 +7,74 @@
...
+    """Tests for the :func:`fedmsg.crypto.utils.fix_datagrepper_message` function."""
+
+    def test_no_source_keys(self):
+        """Assert messages with neither "source_name" or "source_version" are untouched."""
+        original_message = {'my': 'message'}
+
+        self.assertTrue(original_message is utils.fix_datagrepper_message(original_message))
+
+    def test_no_source_name(self):
+        """Assert messages missing the "source_version" key are untouched."""
```
bowlofeggs: Perhaps this comment was meant to say "source_name" instead of "source_version"?

fedmsg/tests/crypto/test_utils.py
```diff
@@ -7,6 +7,74 @@
...
+        original_message = {'source_version': '0.1.0'}
+
+        self.assertTrue(original_message is utils.fix_datagrepper_message(original_message))
+
+    def test_no_source_version(self):
+        """Assert messages missing the "source_name" key are untouched."""
```
bowlofeggs: Similar here.

fedmsg/tests/crypto/test_utils.py
```diff
@@ -7,6 +7,74 @@
...
+        original_message = {'source_name': 'datanommer'}
+
+        self.assertTrue(original_message is utils.fix_datagrepper_message(original_message))
+
+    def test_no_timestamp(self):
+        """Assert messages missing the "source_name" key are untouched."""
```
bowlofeggs: Comment is copypasta'd.

bowlofeggs (approved):

fedmsg/crypto/utils.py
jeremycline: datanommer always adds both, according to the code. My fear was that if a message uses either one of the keys (they're both pretty generic), we'd mess with it. We'll still do the wrong thing if they use both, but this reduces the likelihood.

It's not a great solution. It's not even a _good_ solution, but I don't know what else to do.

fedmsg/crypto/utils.py
jeremycline: We could, and I considered that. The reason I opted not to is because this is already way more complicated than it should be and even if we did that we'd still be wrong plenty of times. What happens when someone sets that to ``null`` rather than an empty object? What happens when messages get signed with floating point timestamps rather than ints? datanommer truncates those. What happens when datanommer adds a new field and forgets to adjust this?

I'd rather have datanommer make strong promises about not altering messages.

jeremycline:
I'm not sure why codecov thinks the diff isn't covered, but it is.

Issue #441, opened 2017-07-05
Use the module name as the logger name


jeremycline:
We should change all fedmsg logging to occur from a logger acquired using ``logging.getLogger(__name__)`` (the recommended practice in the Python [logging documentation](https://docs.python.org/2/library/logging.html#logger-objects)).

There's loggers using the package name ("fedmsg"), a package name outside of fedmsg ("moksha"), and logging instances are being passed around/stored on objects. This makes it much more annoying to track down where logs are being emitted, and also makes configuration unnecessarily difficult.

cha-ku:
Hi @jeremycline , I would like to work on this issue. Could you please assign this to me?

jeremycline:
I don't think I can assign it to you since you're not part of the fedora-infra GitHub organization. I think commenting here is enough to let people know you're working on it.

cha-ku:
Alright then. Will work on this. Thanks! :)

jeremycline:
Fixed in #479

PR #479, opened 2017-09-08
Changed logger to use module name

cha-ku:
Change with reference to Issue #441

### DIFF
--- a/fedmsg/consumers/dummy.py
+++ b/fedmsg/consumers/dummy.py
@@ -38,5 +38,5 @@ class DummyConsumer(FedmsgConsumer):
 
     def consume(self, msg):
         # Do nothing.
-        log = logging.getLogger("moksha.hub")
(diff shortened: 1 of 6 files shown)


fedmsg/consumers/dummy.py
```diff
@@ -38,5 +38,5 @@ def __init__(self, hub):
 
     def consume(self, msg):
         # Do nothing.
-        log = logging.getLogger("moksha.hub")
+        log = logging.getLogger(__name__)
```
bowlofeggs, line 41: This line is missing test coverage.

bowlofeggs (approved):
LGTM, though the coverage.io bot is upset.

jeremycline:
Thanks @chaiku!

Normally I require 100% diff coverage, but in this case I think it's a pretty safe change and the coverage isn't going down.

PR #482, opened 2017-09-28
Alter how the tests determine if cryptography is available

jeremycline:
Rather than re-importing cryptography and m2crypto to detect what tests
to run in the test module, just import from the x509_ng module. This
fixes issues when running the tests with an old pyOpenSSL library that
doesn't provide the APIs we need.


pypingou:
Looks find to me

PR #485, opened 2017-10-10
Add a skeleton for the click-based CLI

jeremycline:


bowlofeggs (approved):
Looks fine to me, though you could easily get coverage by just calling the ```cli()``` function ☺

PR #488, opened 2017-10-26
Add defaults and validators for all valid config keys

jeremycline:
This adds a lazily-loaded dictionary containing all valid fedmsg
configuration keys along with validators for all those keys to ensure
they are set to an acceptable value.

This new configuration dictionary is meant to replace the old
load_config API. That API has been marked as deprecated and will be
removed in a future release.

### DIFF
--- a/fedmsg/config.py
+++ b/fedmsg/config.py
@@ -41,16 +41,464 @@ You can print the runtime configuration to the terminal by using the
 
 import argparse
 import copy
+import logging
 import os
 import sys
 import textwrap
 import warnings
 
+import six
+
 from kitchen.iterutils import iterate
 from fedmsg.encoding import pretty_dumps
 
 
+_log = logging.getLogger(__name__)
+
+
 bare_format = "[%(asctime)s][%(name)10s %(levelname)7s] %(message)s"
+
+
+def _get_config_files():
+    """
+    Load the list of file paths for fedmsg configuration files.
+
+    Returns:
+        list: List of files containing fedmsg configuration.
+    """
+    config_paths = []
+    if os.environ.get('FEDMSG_CONFIG'):
+        config_location = os.environ['FEDMSG_CONFIG']
+    else:
+        config_location = '/etc/fedmsg.d'
+
+    if os.path.isfile(config_location):
+        config_paths.append(config_location)
+    elif os.path.isdir(config_location):
+        # list dir and add valid files
+        possible_config_files = [os.path.join(config_location, p)
+                                 for p in os.listdir(config_location) if p.endswith('.py')]
+        for p in possible_config_files:
+            if os.path.isfile(p):
+                config_paths.append(p)
+    if not config_paths:
+        _log.info('No configuration files found in %s', config_location)
+    return config_paths
+
+
+def _validate_non_negative_int(value):
+    """
+    Assert a value is a non-negative integer.
+
+    Returns:
+        int: A non-negative integer number.
+
+    Raises:
+        ValueError: if the value can't be cast to an integer or is less than 0.
+    """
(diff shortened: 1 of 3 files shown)


bowlofeggs:
It looks like this is failing due to the new version of ```flake8``` that checks for bare ```except``` statements. We had this problem with Bodhi too. You can either add an ignore on ```E722``` to your flake8 config, or you can s/except:/except Exception:/ in the codebase. Looks like there aren't very many, so I'd recommend the latter of the two, and I'd recommend doing it in a separate commit or PR.

fedmsg/config.py
```diff
@@ -41,17 +41,459 @@
...
+    """
+    value = int(value)
+    if value < 0:
+        raise ValueError('Integer must be greater than or equal to zero')
+    return value
+
+
+def _validate_non_negative_float(value):
+    """
+    Assert a value is a non-negative integer.
```
bowlofeggs: s/integer/float/

fedmsg/config.py
```diff
@@ -41,17 +41,459 @@
...
+
+    Returns:
+        float: A non-negative floating point number.
+
+    Raises:
+        ValueError: if the value can't be cast to an integer or is less than 0.
```
bowlofeggs: s/integer/float/

fedmsg/config.py
```diff
@@ -41,17 +41,459 @@
...
+    """
+    value = float(value)
+    if value < 0:
+        raise ValueError('Floating point number must be greater than or equal to zero')
+    return value
+
+
+def _validate_none_or_type(t):
+    """
+    Validate a setting is either None or a given type.
```
bowlofeggs: You might want to express that this actually generates a validator that does this, since it itself does not do this.

fedmsg/config.py
```diff
@@ -41,17 +41,459 @@
...
+
+    Returns:
+        callable: A callable that will validate a setting for that type.
+    """
+    def _validate(setting):
+        """
+        Check the setting to make sure it's the right type.
+
+        Args:
+            object: The setting to check.
```
bowlofeggs: s/object/setting/

fedmsg/config.py
```diff
@@ -41,17 +41,459 @@
...
+        else:
+            raise ValueError('"{}" must be a boolean ("True" or "False")'.format(value))
+
+    if not isinstance(value, bool):
+        raise ValueError('"{}" is not a boolean value.'.format(value))
+
+    return value
+
+
+class FedmsgConfig(dict):
```
bowlofeggs, line 174: Might want a docblock here.

fedmsg/config.py
```diff
@@ -41,17 +41,459 @@
...
+
+        if errors:
+            raise ValueError(
+                'Invalid configuration values were set: \n{}'.format('\n'.join(errors)))
+
+
+#: The fedmsg configuration dictionary. All valid configuration keys are
+#: guaranteed to be in the dictionary and to have a valid value. This dictionary
+#: should not be mutated. This is meant to replace the old :func:`load_config`
+#: API, but is not backwards-compatible with it.
```
bowlofeggs, line 499: You could just use this text as the docblock for the class instead.

bowlofeggs (approved):
This is a fine way to do it (and indeed is how Bodhi does it), but I recently learned that Colander is more generic than I realized and can do stuff like this:

https://docs.pylonsproject.org/projects/colander/en/latest/

Not at all saying you should switch since you already have this working, just thought you might find that library interesting to file away in your soft, human brain.

jeremycline:
Colander has been filed away in my soft human brain for future investigation

PR #489, opened 2017-10-26
PEP8: Remove or ignore bare excepts (flake8 E722)

jeremycline:


bowlofeggs (approved):
Ah I see that you already addressed this issue. This is a fine way to do it, but you can also just s/except:/except Exception:/ without ignoring E722.

jeremycline:
Heh, IIRC I tried that and tests started failing which horrified me. Nearly all that code is on the chopping block in any case so I decided to take the faster route at the moment.

PR #490, opened 2017-11-02
Add a broker service and a publish API

jeremycline:
Hey folks,

This PR consists of two commits:

* Add a broker service that accepts message submissions and publishes them using "publishers".
* Add an API to publish via the broker

I'm interested in all feedback, but what I'm most interested in at the moment is feedback on the publish API. I wanted to get a feel for what people want before I start writing tests and tidying up the code.

To play with this, see the included documentation changes.

CC @ralphbean, @mikebonnet, @bowlofeggs, @pypingou, @abompard 
(I'm sure there are other interested parties I've forgotten about at the moment, sorry!)

### DIFF
--- a/fedmsg/api.py
+++ b/fedmsg/api.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of fedmsg.
+# Copyright (C) 2017 Red Hat, Inc.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+"""The main interface for fedmsg."""
+from __future__ import absolute_import
+import logging
+import json
+
+import six
+import zmq
+
+from . import config, exceptions
+from .broker.service import SUBMISSION_V1
+
+_log = logging.getLogger(__name__)
+
+
+def publish(topic=u'', headers=None, body=None, timeout=10):
+    """
+    Publish a fedmsg via the fedmsg broker service.
+
+    This API blocks until the broker responses with either success or failure.
+
+    Args:
+        topic (str): A unicode string to use as the message topic. If ``topic_prefix``
+            is set it will be prepended to this topic.
+        headers (dict): Optional message headers.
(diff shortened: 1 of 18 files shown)


fedmsg/api.py
```diff
@@ -0,0 +1,99 @@
...
+from . import config, exceptions
+from .broker.service import SUBMISSION_V1
+
+_log = logging.getLogger(__name__)
+
+# Cache connections to the broker in a thread-local
+_thread_local = threading.local()
+
+
+def publish(topic=u'', headers=None, body=None, timeout=10):
```
jeremycline: Do people think it's worth making this as drop-in replaceable for the old publish API as possible? The headers kwarg would need to be at the end and there would need to be some unused kwargs. It's not appealing because it will make it confusing to new users, but it would make transitioning easier, I suppose.

fedmsg/api.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of fedmsg.
+# Copyright (C) 2017 Red Hat, Inc.
```
bowlofeggs, line 4: Since you used (C) instead of ©, do you need the utf-8 declaration?

fedmsg/api.py
bowlofeggs: No strong opinion on this.

fedmsg/broker/publishers.py
```diff
@@ -0,0 +1,153 @@
...
+                               [topic, headers, body], type(self).__name__)
+                result = {'publisher': type(self).__name__, 'result': 'failure', 'reason': str(e)}
+
+            self.pair.send(json.dumps(result).encode('utf-8'))
+
+    def publish(self, topic, headers, body):
+        """
+        This method is called when a message is submitted by a user for publication.
+
+        How the topic, headers, and body are published are up dependent on the type
```
bowlofeggs: s/up//

fedmsg/broker/service.py
```diff
@@ -0,0 +1,174 @@
...
+            if not os.path.exists(socket_dir):
+                _log.info('%s does not exist, creating it for the submission endpoint', socket_dir)
+                os.makedirs(socket_dir, mode=0o775)
+        _log.info('Binding to %s as the fedmsg publisher submission socket',
+                  self.submission_endpoint)
+        self.submission_socket.bind(self.submission_endpoint)
+
+        # Since this method blocks, let users halt us with a signal handler
+        def _handler(signum, frame):
+            """Gracefully signal handler that gracefully shuts down the broker"""
```
bowlofeggs: It'd be good to docblock the args.

bowlofeggs (approved):
Seems fine to me.

pypingou:
Is the broker becoming mandatory or is it a choice when deploying?

jeremycline:
> Is the broker becoming mandatory or is it a choice when deploying?

The old approach will be deprecated and it will eventually be mandatory.

fedmsg/api.py
abompard: I would say keep it as API-compatible as possible and introduce a new API, maybe in an `api2` module?

fedmsg/api.py
abompard: I would say that declaring encoding is always a good thing, even if it's not strictly necessary at the time the file is first written :-)

Issue #455, opened 2017-08-02
Document the replay mechanism and how to stop it


pypingou:
Fedmsg-hub stores on disk in ``/var/run/fedmsg/status/fedmsg-hub/*`` the last fedmsg message received by every consumer it is running. This allows the consumer to go and query datagrepper for any newer messages upon restart.

To stop this behavior (just for one restart), simply nuke this file, fedmsg-hub will not know how long it has to go back and ask datagrepper and thus will not do it.

ralphbean:
Note that you have to 1) stop the hub, 2) rm the file, 3) restart the hub.

If you just nuke it and restart, it might write out another status file right after you nuked it.

ralphbean:
The code is [here](https://github.com/fedora-infra/fedmsg/blob/develop/fedmsg/consumers/__init__.py#L292-L296).

cha-ku:
I'll take this up.

jeremycline:
@chaiku great!

Just send in a PR - I can't assign the issue to users outside the Github organization.

PR #491, opened 2017-11-03
Added documentation for fedmsg-hub

cha-ku:
#455 Added documentation to get all messages from fedmsg-hub after restart.


doc/deployment.rst
```diff
@@ -215,13 +215,23 @@ Start the ``fedmsg-hub`` daemon, which will pick up the datanommer plugin,
 which will in turn read in that connection string, start listening for
 messages, and store them all in the db.
 
+
```
jeremycline: This whitespace change is unnecessary

doc/deployment.rst
```diff
@@ -215,13 +215,23 @@ Start the ``fedmsg-hub`` daemon, which will pick up the datanommer plugin,
...
 ::
 
     $ sudo systemctl start fedmsg-hub
     $ sudo systemctl enable fedmsg-hub
 
 You can check ``journalctl --follow`` for logs.
 
+Since fedmsg-hub stores the timestamp of the last message it had received
```
jeremycline: I think this would fit better in the [subscribing documentation in the consumer section](https://fedmsg.readthedocs.io/en/stable/subscribing/#the-consumer-approach). Add a sub-section called "Message Replay" or something and include this.

Additionally, the first sentence is a fragment sentence. Something like:

"fedmsg-hub stores the timestamp of the last message it has received in a file in the ``/var/run/fedmsg/status/fedmsg-hub/``. It uses this to replay messages it missed from the datagrepper service. To disable this functionality:"

jeremycline (changes requested):
Hey, thanks for the pull request!

I've left a few comments inline. If you could squash your commits and write a commit message that includes ``fixes #455`` as well as a more specific subject line that would be great.

PR #493, opened 2017-11-29
Strengthen "legacy protection".

ralphbean:
Long ago, we introduced a decorator around the msg2 functions that, in the
event of a `KeyError` would try to gracefully return a default value.  This
protection only kicked in *if* the msg2 function was called with
`legacy=True`... i.e., if it was expecting a legacy msg to be handled.

In practice, we never know if a legacy msg is being handled until its too late.
This change renames the handler and broadens its role to catch exceptions from
any message.  They are logged, and the default value is returned.

### DIFF
--- a/fedmsg/meta/__init__.py
+++ b/fedmsg/meta/__init__.py
@@ -158,6 +158,10 @@ def graceful(cls):
     return _wrapper
 
 
+# Deprecated!
+legacy_condition = graceful
+
+
 def with_processor():
     def _wrapper(f):
         def __wrapper(msg, processor=None, **config):
--- a/fedmsg/meta/__init__.py
+++ b/fedmsg/meta/__init__.py
@@ -49,6 +49,8 @@ entry-point, your new class will need to be added to the
 End users can have multiple plugin sets installed simultaneously.
 
 """
+
+import functools
 import logging
 
 import six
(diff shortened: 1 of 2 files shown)


fedmsg/meta/__init__.py
```diff
@@ -135,19 +137,23 @@ def msg2processor(msg, **config):
         return processors[-1]  # DefaultProcessor
 
 
-def legacy_condition(cls):
+def graceful(cls):
```
jeremycline, line 140: Is there any chance that someone's importing this elsewhere? I don't think it's "officially" documented as part of the API, but I don't see a reason to rename the decorator.

fedmsg/meta/__init__.py
```diff
@@ -135,19 +137,23 @@ def msg2processor(msg, **config):
...
+        @functools.wraps(f)
+        def __wrapper(msg, **config):
             try:
                 return f(msg, **config)
-            except KeyError:
-                if legacy:
-                    return cls()
-                else:
-                    raise
+            except Exception:
```
jeremycline: I'd prefer to not handle such a broad exception class here. ``KeyError`` seems appropriate.

If, for example, anyone wants to make an exception class for meta-related problems, this will need to change and people will have started relying (unknowingly) on this. It also looks like this needs test coverage.

jeremycline (changes requested):

ralphbean:
@jeremycline how do you feel about the general approach?

Should I proceed?

Would this (in the end) resolve the problems with fedora-infra/fedmsg_meta_fedora_infrastructure#450?

jeremycline:
I think the general approach is fine, but I'd stick with catching just the KeyError.

fedmsg/meta/__init__.py
ralphbean: Surely there's a chance, but I think it is a low chance.

The `legacy_condition` name is kind of confusing.  "graceful" fits better (as in "graceful degratation").

ralphbean:
OK, this is ready for a new round of review @jeremycline.

fedmsg/meta/__init__.py
jeremycline: I don't have strong opinions about the name, but if you're going to change it,  it wouldn't hurt to add
```
legacy_condition = graceful
```

and document it as deprecated. There's no reason to break the API, even if it's not "official". It can then be removed in a major release.

jeremycline (changes requested):

jeremycline (approved):

PR #497, opened 2017-12-07
Adjust the x509 signing API to return text instead of bytes

### LINKED ISSUE #495: Python 3 incompatibility when signing
After switching to python 3, my code started to crash with:
```
  File "/usr/lib/python3.6/site-packages/koschei/plugins/fedmsg_plugin/backend/publisher.py", line 31, in publish_fedmsg
    fedmsg.publish(**message)
  File "/usr/lib/python3.6/site-packages/fedmsg/__init__.py", line 81, in _wrapper
    return func(*args, **kw)
  File "/usr/lib/python3.6/site-packages/fedmsg/__init__.py", line 93, in publish
    return __local.__context.publish(topic, msg, **kw)
  File "/usr/lib/python3.6/site-packages/fedmsg/core.py", line 328, in publish
    [topic, fedmsg.encoding.dumps(msg).encode('utf-8')],
  File "/usr/lib64/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.6/site-packages/fedmsg/encoding/__init__.py", line 73, in default
    return super(FedMsgEncoder, self).default(obj)
  File "/usr/lib64/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'bytes' is not JSON serializable
```

There are no `bytes` in my message, so I tried to run it with a debugger and print the message right before it fails (long values replaced by `....`):
```python
{'certificate': b'....',
 'crypto': 'x509',
 'i': 1,
 'msg': {'collection': 'f27',
  'collection_name': 'Fedora Rawhide',
  'groups': [],
  'koji_instance': 'primary',
  'name': 'net-snmp',
  'new': 'failing',
  'old': 'ok',
  'repo': 'f27'},
 'msg_id': '2017-065efb41-59ac-4285-abb5-8d7d91e9f260',
 'signature': b'....',
 'timestamp': 1512397293,
 'topic': 'org.fedoraproject.stg.koschei.package.state.change',
 'username': 'koschei'}
```

As you can see, the `certificate` and `signature` keys are of type `bytes`, which makes the json encoder fail.

fedmsg version: `fedmsg-1.0.1-4.fc27.noarch`

CC @mizdebsk

jeremycline:
Instead of returning ascii-encoded bytes, return the signature and
certificate as text so it can be JSON-serialized in Python 3.

fixes #495 

Note: I'm interested what people think about this change. Should it be in a major release? There's an alternate approach to fix this which is a bit hacky, but we can adjust our JSON encoder to check for the signature fields and decode them since we happen to know they should be ASCII. We could do that and defer this for a major release.

### DIFF
--- a/fedmsg/crypto/x509.py
+++ b/fedmsg/crypto/x509.py
@@ -86,8 +86,8 @@ def _m2crypto_sign(message, ssldir=None, certname=None, **config):
     # Return a new dict containing the pairs in the original message as well
     # as the new authn fields.
     return dict(message.items() + [
-        ('signature', signature.encode('base64')),
-        ('certificate', certificate.encode('base64')),
+        ('signature', signature.encode('base64').decode('ascii')),
+        ('certificate', certificate.encode('base64').decode('ascii')),
     ])
 
 
--- a/fedmsg/crypto/x509_ng.py
+++ b/fedmsg/crypto/x509_ng.py
@@ -92,8 +92,8 @@ def sign(message, ssldir=None, certname=None, **config):
         cert_pem = cert.public_bytes(serialization.Encoding.PEM)
 
     return _prep_crypto_msg(dict(list(message.items()) + [
-        ('signature', base64.b64encode(signature)),
-        ('certificate', base64.b64encode(cert_pem)),
+        ('signature', base64.b64encode(signature).decode('ascii')),
+        ('certificate', base64.b64encode(cert_pem).decode('ascii')),
     ]))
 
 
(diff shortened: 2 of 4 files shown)


bowlofeggs (approved):
The change looks good. I'm unsure as to your question of whether this should be a major version or not, as I'm not familiar with this API and who/what uses it. Will fedmsg users (like Bodhi/etc.) have to take some action in response to this change? Can you describe that a bit?

jeremycline:
> Will fedmsg users (like Bodhi/etc.) have to take some action in response to this change? Can you describe that a bit?

They shouldn't. Typically what happens is a user calls ``fedmsg.publish`` which, if the config flag is set, passes the message to this function, and then on to a fedmsg-specific JSON encoder, and then to zmq. No one _should_ be using this API directly, but that doesn't mean they aren't.

I can't think of a good reason for anyone to use the API and the change is quite minor, so I'm on the fence.

bowlofeggs:
If it's not the documented interface then I say don't worry about it.

PR #499, opened 2018-01-08
Feature: Remove dependency on config files

sijis:
This allows a publisher to send a complete configuration via
`fedmsg.init()`.

### DIFF
--- a/fedmsg/__init__.py
+++ b/fedmsg/__init__.py
@@ -53,10 +53,13 @@ def init(**kw):
         raise ValueError("fedmsg already initialized")
 
     # Read config from CLI args and a config file
-    config = fedmsg.config.load_config([], None)
+    config = fedmsg.config.conf
 
-    # Override the defaults with whatever the user explicitly passes in.
-    config.update(kw)
+    settings = {}
+    if kw:
+        settings['settings'] = kw
+
+    config.load_config(**settings)
 
     __local.__context = fedmsg.core.FedMsgContext(**config)
     return __local.__context
--- a/fedmsg/__init__.py
+++ b/fedmsg/__init__.py
@@ -54,12 +54,7 @@ def init(**kw):
 
     # Read config from CLI args and a config file
     config = fedmsg.config.conf
-
-    settings = {}
-    if kw:
-        settings['settings'] = kw
-
-    config.load_config(**settings)
+    config.load_config(kw)
 
     __local.__context = fedmsg.core.FedMsgContext(**config)
     return __local.__context


sijis:
I still need to patch up and create test cases around this change.

I wanted to publish this early enough for feedback.

jeremycline (commented):
I'm on vacation until Wednesday, but at first glance I think this would be better if ``fedmsg.init`` switched to the new ``fedmsg.config.conf`` dictionary which has sane defaults and validators. You'll likely need to remove those validations of the endpoints setting in ``fedmsg.config.load_config``.

sijis:
I did see the deprecation code in `load_config` and I wondered why `init` was still referencing it.

I will convert `init` to use `fedmsg.config.conf`.

sijis:
I'm not sure if this is the correct approach on using `FedmsgConfig()` in the `init()` function.

I still have to poke at test cases around this change but it is somewhat related to #500.

fedmsg/__init__.py
```diff
@@ -53,10 +53,13 @@ def init(**kw):
...
-    config = fedmsg.config.load_config([], None)
+    config = fedmsg.config.conf
 
-    # Override the defaults with whatever the user explicitly passes in.
-    config.update(kw)
+    settings = {}
+    if kw:
+        settings['settings'] = kw
+
+    config.load_config(**settings)
```
jeremycline: Rather than creating a dict with a single key and immediately unpacking it, just do ``config.load_config(kw)``.

jeremycline (changes requested):

fedmsg/__init__.py
sijis: I was waffling back and forth on how to do it. I figured there was a simpler way :). Thanks.

jeremycline:
Could you rebase this onto the latest master? That should take care of the test suite failures.

sijis:
Rebased.

jeremycline (changes requested):
It looks like this breaks a couple of the tests on Python 2.7 - the validator expects unicode, but it's getting strings.

The fix should be to go to those failing tests which are using invalid config and make sure those strings are prefixed with ``u``. Thanks!

sijis:
I've been poking at this for over an hour and I think the issue is more than just prefixing those strings with `u`.

There's definitely a discrepancy between `fedmsg.config.conf` and using the old `load_config()`.

I started poking around and seeing if i could patch https://github.com/fedora-infra/fedmsg/blob/develop/fedmsg/commands/__init__.py#L45, so I could fix the `test_logger_basic()` test but that's leading me down a rabbit hole.

jeremycline:
I poked at this a little bit and ooof, this stuff is messy. The current CLI is pretty awful and should just be tossed out, as should the old config setup... Really the whole fedmsg API.

Honestly, I feel that it's probably not worth trying to get this to work. I don't know what exactly you're using fedmsg for, but if it's not to work with existing Fedora infrastructure I _really_ think you'd be happier using plain ZMQ. It's super simple to use and fedmsg is likely just getting in your way.

That's not to say I won't accept this PR if you can make the current CLI and tests work with the new configuration, but I'm not sure it's the best use of time.

PR #500, opened 2018-01-08
Fix default type check

sijis:
While working on #499, I noticed some discrepancies in the validation of the configs in `FedmsgConfig`.

This attempts to address those discrepancies.

I used the generated output from `fedmsg-config` as the source of truth on what these values should be.


fedmsg/config.py
```diff
@@ -254,8 +254,10 @@ class FedmsgConfig(dict):
             'validator': None,
         },
         'relay_inbound': {
-            'default': u'tcp://127.0.0.1:2001',
-            'validator': _validate_none_or_type(six.text_type),
```
jeremycline: It doesn't make much sense to have a relay binding to multiple sockets. Did you run into this by trying to specify a list of values?

sijis: I did not run this using multiple entries.

I think the issue is mainly documentation related. 
* The existing config show that parameter as a list (https://github.com/fedora-infra/fedmsg/blob/develop/fedmsg.d/relay.py#L36)
* The `fedmsg-config` command also shows this option as a list. 
* The online docs reference it as a list too (http://fedmsg.readthedocs.io/en/stable/configuration/#relay-inbound)

In my case, I just updated the example config and things worked fine. I never needed to add multiple values to this.

It is very likely I am wrong and reading the wrong documentation parts.

jeremycline: Ah, well the history here is I wrote documentation. There wasn't any documentation before and I took wild guesses. Recently I tried to add defaults to every setting and picked what I thought were reasonable defaults.

Although it looks like lists and strings are both handled correctly, it's not a setting that makes a bunch of sense as a list so I'm in favor of changing the documentation and config in ``fedmsg.d`` (really, that directory should go away, but I have no doubt that'd break something weird relying on it 😢 )

sijis: I will definitely change/revert that list change in this PR.
I could also include documentation changes so that it references it as a string, not list. This would keep everything consistent, at least.

At quick check, tests fail removing the `fedmsg.d` directory. I bet packaging would have issues too. Although, both of those could be solved.

sijis: I've removed this commit completely.

sijis:
Updated references of `relay_inbound` to be strings.

fedmsg/config.py
```diff
@@ -294,8 +294,8 @@ class FedmsgConfig(dict):
             'validator': _validate_none_or_type(six.text_type),
         },
         'crl_cache_expiry': {
-            'default': None,
-            'validator': _validate_none_or_type(six.text_type),
+            'default': 10,
```
jeremycline: The docs say this is in seconds (although the docs _are_  a little unreliable) so 10 is a really small default for a CRL. Maybe an hour is more reasonable?

jeremycline (changes requested):
Sorry, I should have made an note on the other changes on the first review. One small thing, but other than that it looks really good, thanks!

fedmsg/config.py
sijis: Good point. Updating...

jeremycline: I think you missed updating it in a few places? Additionally, an hour would be 60*60 (3600) since this is seconds, not minutes.

sijis: You don't want to to expire in 1 minute? 😂 
Let me try this again.

sijis: Totally missed several spots.
Alright I think it finally got this.
Also, sorry for the constant rebasing. Just trying to keep clean commits.

jeremycline (approved):

jeremycline:
Great, thanks for this!

PR #507, opened 2018-03-09
Return any return value from ._consume(..).

ralphbean:
In a non-zeromq case, it matters what the return value of _consume is.  (It is
a boolean, indicating if the message was successfully handled or not).  This,
in turn, gets used to figure out whether or not to send an ACK back to the
broker.

Here, the return values were being swallowed and turned into `None` values.

### DIFF
--- a/fedmsg/consumers/__init__.py
+++ b/fedmsg/consumers/__init__.py
@@ -290,11 +290,11 @@ class FedmsgConsumer(moksha.hub.api.consumer.Consumer):
 
                 try:
                     self.validate(m)
-                    super(FedmsgConsumer, self)._consume(m)
+                    return super(FedmsgConsumer, self)._consume(m)
                 except RuntimeWarning as e:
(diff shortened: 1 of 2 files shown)


mikebonnet:
:+1:

jeremycline (changes requested):
The change itself looks reasonable. Can you please write documentation for the function including its return value and tests?

Thanks.

ralphbean:
Docs and tests should be there, now.

jeremycline (approved):

jeremycline:
Thanks!

ralphbean:
Yay :)

@jeremycline, mind if I prepare a `fedmsg-1.1.1` release?

jeremycline:
Ah, sorry I missed this. For future reference I don't mind, but since I'm here I'll just write the release notes and tag a release.

ralphbean:
Hey, thanks!

PR #513, opened 2018-04-19
WIP: Implement a message API with schemas

jeremycline:
This adds a messages module that introduces a new entry point for
fedmsg. This entry point is used to register new message classes to
topics that include schema definitions.

This is super rough, no tests, etc. Basic usage would be like this:

1. Make a small Python package with your message class (defining a valid jsonschema for headers/body) and in its setup.py include an entry point:
```
entry_points = {'fedmsg.messages': ['my.topic=mypackage.mymodule:MyMessageClass']}
```

2. Install this package on hosts that wish to send/receive that message.

This then would get hooked up to the publish/subscribe API and validate would be called before sending/after receiving.

@bowlofeggs

### DIFF
--- a/fedmsg/message.py
+++ b/fedmsg/message.py
@@ -1,0 +1,214 @@
+"""
+This module defines the base class of message objects and keeps a registry of known
+message implementations. This registry is populated from Python entry points. The registry
+keys are message topics and the values are sub-classes of :class:`Message`.
+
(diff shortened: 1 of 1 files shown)


bowlofeggs:
@jeremycline is this obsoleted by the new Fedora messaging stuff?

jeremycline:
Yerp

PR #514, opened 2018-05-22
handle messages with non-dict bodies

mikebonnet:
Lots of places in the fedmsg code assume that message['body'] is a dict. However,
it is possible for a producer to send a message with a body that is valid JSON
but does not encode a dict, which would cause validate() to fail. This change
ensures that message['body'] is always a dict.

### DIFF
--- a/fedmsg/consumers/__init__.py
+++ b/fedmsg/consumers/__init__.py
@@ -252,7 +252,7 @@ class FedmsgConsumer(moksha.hub.api.consumer.Consumer):
                 message['body'] = json.loads(message['body'].decode('utf-8'))
 
         # Massage STOMP messages into a more compatible format.
-        if 'topic' not in message['body']:
+        if not isinstance(message['body'], dict) or 'topic' not in message['body']:
             message['body'] = {
                 'topic': message.get('topic'),
                 'msg': message['body'],
--- a/fedmsg/tests/consumers/test_consumers.py
+++ b/fedmsg/tests/consumers/test_consumers.py
@@ -110,6 +110,22 @@ class FedmsgConsumerValidateTests(unittest.TestCase):
         self.consumer.validate(message)
         self.assertEqual({'body': {'topic': None, 'msg': {'some': 'stuff'}}}, message)
 
+    def test_none_body(self):
+        """Assert that a message body of None is replaced with an empty dict."""
+        self.consumer.validate_signatures = False
+        message = {'body': None}
+
+        self.consumer.validate(message)
+        self.assertEqual({'body': {'topic': None, 'msg': None}}, message)
+
+    def test_string_body(self):
+        """Assert that a message body that is a string is replaced with a dict."""
+        self.consumer.validate_signatures = False
+        message = {'body': 'this is the body'}
+
+        self.consumer.validate(message)
+        self.assertEqual({'body': {'topic': None, 'msg': 'this is the body'}}, message)
+
     @mock.patch('fedmsg.consumers.fedmsg.crypto.validate')
     def test_zmqmessage_text_body(self, mock_crypto_validate):
         self.consumer.validate_signatures = True


mikebonnet:
@jeremycline @pypingou @ralphbean Mind taking a look?

jeremycline:
Rather than allowing publishers to publish messages that are malformed*, wouldn't it be better to raise an exception when they try to publish the message. I don't like the magical wrapping going on the consumer, it makes trying to figure out what the message is going to look like at any given point even crazier than it already is.

* Since there's no schema I guess maybe it's not "malformed", exactly. Are there publishers in the wild doing this?

Also, @abompard is the maintainer now.

mikebonnet:
This isn't adding any new magic wrapping, just reinforcing the wrapping that's already happening. Without this, you get an ugly traceback when the code assumes that `message['body']` is iterable. If we want to enforce that clients always send messages with dict bodies, there should probably be an explicit check for it.

Not seeing this on fedmsg, but we are seeing it on the internal bus. With this change, we could at least see the malformed messages in datagrepper and request that they be fixed. As it is, they cause a traceback and are essentially lost.

jeremycline:
> This isn't adding any new magic wrapping, just reinforcing the wrapping that's already happening. Without this, you get an ugly traceback when the code assumes that message['body'] is iterable. If we want to enforce that clients always send messages with dict bodies, there should probably be an explicit check for it.

Yeah, that's what I'm saying. It's (very unfortunately) implicitly assumed throughout Fedora infrastructure that message bodies are dictionaries. Rather than hiding the problem by dropping it in a dictionary in the consumer, we shouldn't let anyone even publish a message that violates this implicit schema.

Mutating the message creates an awful user experience because you send one thing and receive something else entirely, and that's really confusing. It also causes lots of problems with things like message signatures (datagrepper _already_ mangles the message by inserting keys which breaks the signature and required a hacky work-around in fedmsg). In fact, I'm pretty sure this change breaks message signature validation for those malformed messages.

mikebonnet:
If the message body isn't a dict, message signature validation is moot, since there's no signature. I don't love all the implicit wrapping either, but in this case I think it's better to try to deliver something, rather than just tracing back. In any case, this isn't adding new wrapping, just adding an extra case to the existing wrapping logic.

jeremycline:
Sure, internally you may not be using signatures (I don't know), but what happens if someone does this in Fedora? It makes no sense to me fix it on the consumer side. We should handle it, certainly, by logging very loudly that message x on topic y is being discarded because someone is doing something they shouldn't. The real problem is that fedmsg lets you publish this at all.

Are these messages not being sent by fedmsg, is that why you want to fix it consumer-side?

mikebonnet:
@jeremycline That's correct, we have messages being published by many services, in a variety of languages, and using different frameworks and protocols. We are not using zeromq at all, it's all AMQP and STOMP. We are trying to encourage all producers to use text/json bodies, but it is not feasible for all producers to use fedmsg.

jeremycline:
Ah, that makes a bit more sense. It's helpful to know the context.

Well, you might be pleased to hear we're trying to get [Fedora to move to AMQP](https://lists.fedoraproject.org/archives/list/infrastructure@lists.fedoraproject.org/thread/HLOYNCI4X6ELF76HP54UBMMNL4FPDBQW/). It'd be good if you weighed in with what you'd want to make it easy to re-use as much as possible. For example, we're using jsonschema to define messages schema, which should be language independent.

For this PR, I'd say that it should have a check in the fedmsg publisher so that user's have to really work to shoot themselves in the foot. I don't like mutating the message, but if it's a code path fedmsg users can't hit (and includes loud logging letting them know they've done something crazy) I guess that's okay.

mikebonnet:
@jeremycline As mentioned in the comment on line 254, this code path is only hit by STOMP consumers.

jeremycline:
It's hit by any consumer if the publisher sends a non-dict message. The comment is flat out wrong because it depends on the publisher behavior which we can't control because, as you mentioned, people are publishing from multiple languages using multiple frameworks.

Even with just fedmsg, the publish API lets me send this:
```
fedmsg.publish('hi', 'hi')
```

Which arrives as:
```
[
    b'org.fedoraproject.dev.__main__.hi',
    b'{"i":3,"msg":"hi","msg_id":"2018-e5b3b59f-374f-40ee-83b9-b1fc3bc67131","timestamp":1527779780,"topic":"org.fedoraproject.dev.__main__.hi","username":"jcline"}'
]
```

I'm not going to dig into moksha/fedmsg to see if it's adding a 'body' key somehow and mangling the message to make it work. I'm pretty dubious that this works end-to-end. Since the plan is to abandon fedmsg it doesn't matter much to me, though. I'm not a maintainer anymore, either, so what I think isn't really important in any case.

hluk:
Ah, why is this not merged? We just hit the problem on prod server and I created another PR for this: #523

Any chance for this to get merged and released? Otherwise, we will implement workaround in our consumer.

**Edit:** Actually, I'm not sure if we can easily fix this in our consumer.

ralphbean (approved):

ralphbean (approved):

PR #515, opened 2018-05-23
always enable the MonitoringProducer

mikebonnet:
moksha records the time it takes a Consumer to process every message
in the Consumer._times list. The MonitoringProducer sends those times
to the monitor socket, and resets the list. If the MonitoringProducer
isn't running, that list grows forever, causing a memory leak. The
default configuration should always enable the MonitoringProducer.

### DIFF
--- a/fedmsg.d/monitoring.py
+++ b/fedmsg.d/monitoring.py
@@ -1,0 +1,7 @@
+# Enable the MonitoringProducer.
+# Without this, consumers will leak memory.
+
+config = {
+    'moksha.monitoring.socket': 'ipc:///var/tmp/moksha-monitor',
+    'moksha.monitoring.socket.mode': '777',
+}
--- a/fedmsg/tests/test_commands.py
+++ b/fedmsg/tests/test_commands.py
@@ -341,8 +341,10 @@ class CheckTests(unittest.TestCase):
         thread = threading.Thread(target=report_in_thread)
         thread.start()
 
-    def test_no_monitor_endpoint(self):
+    @mock.patch('fedmsg.commands.check.load_config')
+    def test_no_monitor_endpoint(self, mock_load_config):
         """Assert that when no endpoint for monitoring is configured, users are informed."""
+        mock_load_config.return_value = {}
         expected_error = (
             u'Error: No monitoring endpoint has been configured: please set '
             u'"moksha.monitoring.socket"\n'


fedmsg.d/monitoring.py
```diff
@@ -0,0 +1,7 @@
+# Enable the MonitoringProducer.
+# Without this, consumers will leak memory.
+
+config = {
+    'moksha.monitoring.socket': 'ipc:///var/tmp/moksha-monitor',
```
mikebonnet, line 5: I'm open to changing this to `/var/tmp/fedmsg-monitor` if people think that's more appropriate.

mikebonnet:
@jeremycline @pypingou @ralphbean Mind taking a look?

jeremycline (commented):
If you're changing the default config the change belongs in fedmsg/config.py. However, this seems like a problem that needs to get fixed in moksha, probably by realizing there's no producer and not recording times. If someone turns off the monitor, they're back to leaking memory.

mikebonnet:
@jeremycline I thought about fixing this in moksha, but that would require introspecting the hub and looking for a producer which is an instance of `MonitoringProducer`, which seemed expensive and racy. Two lines of config looked preferable. I'm fine moving them to `config.py` if that's a better place for them.

jeremycline:
You're not actually fixing the problem, though, just hiding it behind a configuration default (a configuration, I might add, for a different project) that someone might change. moksha should _not_ be dumping stuff in a list forever and ever hoping something is removing them. Make it a bounded queue and just drop times if you don't want to deal with the introspection.

mikebonnet:
I think you could argue that this is a configuration problem in fedmsg. The design of moksha essentially requires that *something* reads/resets those stats at regular intervals, but the default configuration bundled with fedmsg does not enable MonitoringProducer (or something else that performs a similar function).

jeremycline:
No, I don't think that's something you could reasonably argue. For one thing, moksha doesn't state that that is a requirement at all. The library has a list in a private variable of one of its classes that, in a perfectly valid configuration*, leaks memory. That's a bug in moksha. It may be the design of moksha, but it's also wrong. If we added it as a default configuration option fedmsg would also be wrong because we'd let a user configure it such that it leaks memory. Let's fix it where it's broken.

*I _assume_ it's valid, although it's not [documented anywhere that you must set this configuration, or that it even is a configuration](https://moksha.readthedocs.io/en/latest/search/?q=moksha.monitoring.socket&check_keywords=yes&area=default)

mikebonnet:
@jeremycline Fair enough, see https://github.com/mokshaproject/moksha/pull/57

PR #520, opened 2018-09-13
Fix fedmsg-hub --with-consumers option

larchunix:
Use explicit_hub_consumers list to build consumers list forwarded to
moksha.hub. Up until now, consumers list was always None.


ralphbean (approved):
Nice!

Zlopez (approved):

larchunix:
@ralphbean, @Zlopez: ping, is there anything I must do for this PR to be merged?

Zlopez:
@abompard I think you are now the maintainer of fedmsg. So it should be your call for merging.

PR #522, opened 2019-12-02
Allow the maximum number of open files for fedmsg-hub to be configurable

mprahl:
fedmsg-hub hard-coded a maximum number of open files to 4096. This
was unexpected behavior when the limit was raised in the systemd
configuration but the limit of 4096 was still reached. This commit
allows the number of open files to be configurable while maintaining
backwards compatibility.

### DIFF
--- a/fedmsg/commands/hub.py
+++ b/fedmsg/commands/hub.py
@@ -24,13 +24,6 @@ from fedmsg.utils import load_class
 from fedmsg.commands import BaseCommand
 
 
-# Since many services use fedmsg-hubs, processes can exceed their resource
-# limits on the number of open file descriptors. In the long term, apps should
-# stop listening to every service, but as a short term fix an attempt is made
-# to raise the resource limit here.
-MAX_NOFILE = 4096
-
-
 class HubCommand(BaseCommand):
     """ Run the fedmsg hub.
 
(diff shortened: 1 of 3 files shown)


mprahl:
@jeremycline or @ralphbean could you please take a look?

jeremycline:
I no longer do any Fedora infrastructure work so I'm not going to make a release for this. If someone on the infra team gives you permissions to do so that's fine by me, but I consider fedmsg dead. As for the change itself it looks fine, although I'd probably not bother even keeping backwards compatibility here since the service almost certainly shouldn't be managing it's own limits.

glensc:
@jeremycline what is the replacement if it's dead? could this be put out somewhere more visible? a pinned ticket in github? readme section?

jeremycline:
@glensc the readme has a note about fedora-messaging, but any AMQP client you want to use should work with the public broker documented at https://fedora-messaging.readthedocs.io/en/latest/quick-start.html#fedora-s-public-broker

glensc:
@jeremycline I have few packages written for fedmsg, running in our infra (not fedora related at all). how to convert them to fedora-messaging, is there guide? where to start?

- https://github.com/glensc/fedmsg-cvs
- https://github.com/glensc/fedmsg-cvs2jira
- https://github.com/glensc/fedmsg-cvs2slack

jeremycline:
https://fedora-messaging.readthedocs.io/en/stable/tutorial/conversion.html would probably be a good place to start. I also recommend just reading the entire [user guide](https://fedora-messaging.readthedocs.io/en/stable/#user-guide).

PR #530, opened 2022-06-09
Fix compatiblity with Python 3.11+

hroncok:
AttributeError: module 'inspect' has no attribute 'getargspec'

See from https://docs.python.org/3.11/whatsnew/3.11.html#removed

Removed from the inspect module:

> the getargspec function, deprecated since Python 3.0; use inspect.signature() or inspect.getfullargspec() instead.

Downstream bug report: https://bugzilla.redhat.com/show_bug.cgi?id=2095027

PR #531, opened 2022-11-16
replace usage of inspect.getargspec() which was removed in Python 3.11

mikebonnet:
`getargspec()` has been deprecated since Python 3.0. `getfullargspec()` provides a compatible API.


mikebonnet:
Duplicate of #530.

PR #532, opened 2023-01-30
Fix the Invalid argument error

Zlopez:
This change will fix the error happening with zmq >= 23. `hasattr` still returned true even if the `HWM` can't be used anymore in zeromq 3+. This will rather check, if the version string for zeromq starts with `2.` More details about this issue could be found in https://pagure.io/fedora-infrastructure/issue/11023


nirik:
Travis? yikes... I will see if I can disable that. We should setup other ci here.

nirik (approved):
I wonder if this is confused because I removed travisci while it was trying to use it?

Not sure, but anyhow.. +1!

PR #533, opened 2023-02-10
Fix tests

Zlopez:
This fixes tests for fedmsg.

* Fixes docs tox env
* Fixes lint errors

Signed-off-by: Michal Konečný <mkonecny@redhat.com>

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -73,7 +73,7 @@ release = u'1.1.2'
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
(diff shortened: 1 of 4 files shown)


Zlopez:
I wasn't able to fix python 3.6 and python 2.7 tests.
Python 2.7 tests are failing on import error, which I couldn't figure out how to fix
```
tox/py27/lib/python2.7/site-packages/py/_path/local.py:704: in pyimport
    __import__(modname)
fedmsg/__init__.py:91: in <module>                                                                                     
    @API_function(doc=fedmsg.core.FedMsgContext.publish.__doc__)   
E   AttributeError: 'module' object has no attribute 'core'        
```
Python 3.6 is failing on very strange errors (empty body of messages), but because the production machine is RHEL 7 with python 2.7.5 I didn't even tried to fix them, so I don't break something on python 2.7, which I couldn't test

PR #534, opened 2023-02-21
Add configuration for skipping last message

penguinpee:
This adds 'skip_last_message' config option, which will skip the message fedmsg-hub worked on last upon retrieval of the queue from datagrepper.

The idea is to have the processing advance at the expense of 'losing' a badge.

### DIFF
--- a/fedmsg/consumers/__init__.py
+++ b/fedmsg/consumers/__init__.py
@@ -141,6 +141,7 @@ class FedmsgConsumer(moksha.hub.api.consumer.Consumer):
                 os.makedirs(topmost_directory)
 
         self.datagrepper_url = self.hub.config.get('datagrepper_url')
+        self.skip_last_message = self.hub.config.get('skip_last_message')
         if self.status_filename and self.datagrepper_url:
             # First, try to read in the status from a previous run and fire off
             # a thread to set up our workload.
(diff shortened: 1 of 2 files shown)


fedmsg/consumers/__init__.py
```diff
@@ -190,7 +191,10 @@ def _backlog(self, data):
 
             if message['msg_id'] != last['msg_id']:
                 retrieved = retrieved + 1
-                self.incoming.put(dict(body=message, topic=message['topic']))
+                if (self.skip_last_message and retrieved > 1):
+                    self.incoming.put(dict(body=message, topic=message['topic']))
+                else:
```
Zlopez, line 196: I'm not sure if this will work as intended, it will just skip every message if the `self.skip_last_message` is set to `false`, because the `if` statement  will be true only if `self.skip_last_message` is set to `true` and `retrieved > 1`, everything else will just skip messages. Which I think is not what you want with this PR.

penguinpee: Thanks for reviewing. I pledge tunnel vision. I was focused on getting messages dropped. What you say is true. Luckily, _skip_last_message_ is currently set to _True_. :smile: 

Will fix and update this PR and the live environment.

penguinpee:
I fixed the logic. It should now behave correctly when `skip_last_message` is _not_ set or set to `False`.

Zlopez (approved):
It looks good now

Zlopez:
I will probably release a new version soon, because I'm trying to fix the build on EPEL7. So this will be part of that version.

PR #539, opened 2023-07-03
Replace the usage of assertEquals unit test alias removed in Python 3.12 with assertEqual

hrnciar:


mikebonnet:
Could we get this merged?

